PR #71, opened 2017-12-20
Add uninstall CLI option

berndtj:
* add a new uninstall command
  - works just like install, but in the reverse
* use a ~/.dispatch directory for CLI config instead of ~/.dispatch.json
  file
  - contains config.json and certs
* add validation to install config file (jsonschema)
  - uninstall does not validate

### DIFF
--- a/pkg/dispatchcli/cmd/cmd.go
+++ b/pkg/dispatchcli/cmd/cmd.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


pkg/dispatchcli/cmd/cmd.go
```diff
@@ -35,6 +36,7 @@ var validResources = i18n.T(`Valid resource types include:
 	* install
 	* secrets
 	* subscription
+	* uninstall
```
kars7e: I think install and uninstall do not belong here, actually this whole variable could be removed. The idea for this variable was to generalize `get`, but since we build subcommand for every type of resource, `dispatch get --help` prints the list anyway. Seems like this variable is not needed.

pkg/dispatchcli/cmd/install.go
```diff
@@ -17,58 +18,159 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/go-openapi/spec"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/validate"
+	"github.com/imdario/mergo"
+	homedir "github.com/mitchellh/go-homedir"
```
kars7e: nit: this alias is not needed

PR #73, opened 2017-12-21
Add e2e pipeline

kars7e:
Pipeline can be found here: https://ci.dispatchframework.io/teams/main/pipelines/dispatch-e2e
**Note**: pipeline is able to execute e2e now, however, there are failures in the tests themselves. we need to fix bugs / verify in case it's environment specific.

* Adds a pipeline that executes for every PR that has `run-e2e` label added.
* Updated docker images to use `vmware/*` versions.
* Simplified unit tasks (removed scripts and moved their content inline)


ci/e2e/binaries.yml
```diff
@@ -0,0 +1,27 @@
+---
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: kars7e/photon-golang-ci
```
berndtj: Let's get this in vmware org

ci/e2e/cleanup.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: dispatch-cli
+
+run:
+  path: /bin/bash
+  args:
+  - -c
+  - |
+    set -e -x -u
+    source dispatch/ci/e2e/config-k8s-env.sh
+    helm delete --purge dispatch api-gateway openfaas ingress || true
```
berndtj: we could also use `dispatch uninstall`

ci/e2e/config-dispatch.sh
```diff
@@ -0,0 +1,37 @@
...
+    export IMAGE_TAG=$(cat properties/keyval.properties | sed 's/tag=//')
+fi
+export DOCKER_REGISTRY_HOST=$(cat cluster/metadata | jq -r '.registryHost')
+export DOCKER_REGISTRY_USER=$(cat cluster/metadata | jq -r '.registryUser')
+export DOCKER_REGISTRY_PASS=$(cat cluster/metadata | jq -r '.registryPass')
+export DOCKER_REGISTRY_EMAIL=$(cat cluster/metadata | jq -r '.registryEmail')
+
+echo "${NODE_IP} dev.dispatch.vmware.com api.dev.dispatch.vmware.com" >> /etc/hosts
+
+cat << EOF > config.yaml
```
berndtj: This file can be output via `dispatch install -d <path to config.yaml>`.  That will write the config file

ci/e2e/binaries.yml
kars7e: will do, asked Mark to create repos.

ci/e2e/cleanup.yml
kars7e: I was using it originally but needed more control what is deleted during testing. Also, verbosity would be useful but debug prints config content which contains sensitive credentials. We can change it once the pipeline gets stable.

ci/e2e/config-dispatch.sh
kars7e: This is the installation input config, did you mean the CLI config stored by default in .dispatch, and which is created in `run-tests` step? if so, there are two problems with it:
* passing data between jobs in Concourse is surprisingly difficult, you need an external resource (like S3) to store file. It's just easier to create this file from the script.
* install does not set "skipAuth" flag, which is needed here.

berndtj: It's fine we can add skipAuth later if we need.  That sucks about not being able to pass artifacts in concourse.

berndtj:
I'm giving you the approved as it looks good, but I see it has the WIP tag.  You can do what you like :).

kars7e:
I removed the WIP. I will merge it so it's there, although e2e will report failures for now until we fix these bugs (we will be able to merge with e2e failure though).

PR #78, opened 2017-12-22
add postgres database

seanhuxy:
- add postgres as entity store backend
- removed postgres chart in kong
- refactor the install.go, the config.yaml is more clear now
- add config.template.yaml so that people can use the default value ASAP.

- note: remove the default config and schema validation in install.go for quick dev. config.template.yaml is sufficient to replace the hard-code default config. and we can add schema back easily if the config.yaml is stable. (also suggest: move the shema to a separate file)


Gopkg.lock
```diff
@@ -44,10 +44,16 @@
   revision = "bcc4c8345a21301bf47c032ff42dd1aae2fe3027"
 
 [[projects]]
-  branch = "master"
-  name = "github.com/coreos/go-oidc"
-  packages = ["."]
-  revision = "a4973d9a4225417aecf5d450a9522f00c1f7130f"
+  name = "github.com/coreos/etcd"
```
kars7e: Is this some leftover? Or you added both postgres and etcd support?

charts/dispatch/charts/event-manager/templates/deployment.yaml
```diff
@@ -31,7 +31,11 @@ spec:
             - "--organization=$(ORGANIZATION)"
             - "--host=0.0.0.0"
             - "--port={{ .Values.service.internalPort }}"
-            - "--db-file=/data/{{ template "name" . }}/db.bolt"
+            - "--db-file={{ default .Release.Name .Values.global.db.release }}-{{ .Values.global.db.host }}.{{ default .Release.Namespace .Values.global.db.namespace }}:{{ .Values.global.db.port }}"
+            - "--db-backend={{ .Values.global.db.backend }}"
```
kars7e: This adds a lot of flags, it would be easier, more concise and more flexible to just pass dbURL in a form of: `dbType://USERNAME:PASSWORD@HOSTNAME/DATABASE`. Golang pq library has an [example of it](https://godoc.org/github.com/lib/pq).

Gopkg.lock
```diff
@@ -336,18 +378,18 @@
...
-
 [[projects]]
   name = "github.com/satori/go.uuid"
   packages = ["."]
   revision = "879c5887cd475cd7864858769793b2ceb0d44feb"
   version = "v1.1.0"
 
+[[projects]]
+  branch = "master"
+  name = "github.com/seanhuxy/libkv"
```
kars7e: I see you added your fork as a dependency. I'm totally ok with reusing some of our code from external repos, but since you replaced the original repo with your fork, you would have to maintain your fork and pull patches, resolve merge conflicts, etc. Based on my brief research on docker/libkv, I doubt they will be interested in a PR for postgres support.

I would suggest implementing it in dispatch entitystore instead. we already provide an abstraction, we just need different implementation based on postgres. this way you won't need to use libkv at all, and you can put your code directly in vmware/dispatch.

Gopkg.lock
```diff
@@ -240,6 +264,18 @@
   revision = "96dc06278ce32a0e9d957d590bb987c81ee66407"
   version = "v1.3.0"
 
+[[projects]]
+  name = "github.com/jinzhu/gorm"
```
kars7e: Our current entitystore implementation has no relationship model. There should be no need for an ORM library. Did you find using `lib/pq` directly insufficient? 
BTW, postgres has support for hstore and JSONB database, which might be a good fit in this use case.

pkg/dispatchcli/cmd/install.go
```diff
@@ -18,54 +18,89 @@ import (
 	"strings"
```
kars7e: This file seems unfinished (a lot of commented code, removed libraries that are actually needed, etc.).

Gopkg.lock
berndtj: This was my idea.  I see your point @kars7e.  I haven't researched docker/libkv deeply.  I'm not sure why they wouldn't accept a Postgres backend, but since we do have our own abstraction, I guess it could work either way.  I don't have a really strong opinion on this.

charts/dispatch/charts/event-manager/templates/deployment.yaml
berndtj: 👍

pkg/dispatchcli/cmd/uninstall.go
```diff
@@ -159,35 +158,41 @@ func runUninstall(out, errOut io.Writer, cmd *cobra.Command, args []string) erro
...
-		err = helmUninstall(out, errOut, "kube-system", "ingress", false)
+		err = helmUninstall(out, errOut, config.Ingress.Chart.Namespace, config.Ingress.Chart.Release, false)
+		if err != nil {
+			return errors.Wrapf(err, "Error uninstalling nginx-ingress chart")
+		}
+	}
+	if uninstallService("postgres") {
+		err = helmUninstall(out, errOut, config.PostgresConfig.Chart.Namespace, config.PostgresConfig.Chart.Release, false)
 		if err != nil {
 			return errors.Wrapf(err, "Error uninstalling nginx-ingress chart")
```
berndtj: This comment should now read Postgres

pkg/dispatchcli/cmd/uninstall.go
```diff
@@ -35,10 +34,10 @@ var (
 	uninstallNamespace       = i18n.T(``)
 	uninstallRemoveCertFiles = false
 
-	defaultUninstallConfig = installConfig{
-		Namespace: "dispatch",
-		Hostname:  "dispatch.vmware.com",
-	}
+	// defaultUninstallConfig = installConfig{
```
berndtj: So is the config a required arg now?

pkg/dispatchcli/cmd/uninstall.go
```diff
@@ -146,10 +145,10 @@ func runUninstall(out, errOut io.Writer, cmd *cobra.Command, args []string) erro
 		return errors.Wrapf(err, "Error decoding yaml file %s", uninstallConfigFile)
 	}
 
-	err = mergo.Merge(&config, defaultUninstallConfig)
-	if err != nil {
-		return errors.Wrapf(err, "Error merging default values")
-	}
+	// err = mergo.Merge(&config, defaultUninstallConfig)
```
berndtj: don't comment out, just remove if we don't need it.

config.template.yaml
```diff
@@ -0,0 +1,58 @@
+HelmRepositoryURL: https://s3-us-west-2.amazonaws.com/dispatch-charts
```
berndtj: There is a LOT more going into config.yaml now.  I was really intending to keep it as small as possible.  Also, I don't see a lot of these new values actually being used.  There certainly aren't any changes to install.go.

Gopkg.lock
berndtj: Let's discuss this a little more.  Certainly we could use JSONB or HSTORE, but we lose some of the query ability and performance of postgres.  Do we want to keep looking like a k/v store, or embrace the SQL?

kars7e: To explain my suggestion of JSONB, I was referring to the current implementation in seanhuxy/libkv, which looks like this:

```go
type KV struct {
 	Key       string `gorm:"primary_key"`
 	Value     []byte
 	UpdatedAt time.Time
 	CreatedAt time.Time
 }
```
and stores the value in `[]byte`, which is mapped to Postgres binary type `bytea` (https://github.com/jinzhu/gorm/blob/b1885a643b4977c9089d77eb07c0fd96591f94b8/dialect_postgres.go#L70)

If we use JSONB, we actually gain query ability and performance, while still using SQL:

DDL:
```sql
CREATE TABLE cards (
  id integer NOT NULL,
  board_id integer NOT NULL,
  data jsonb
);
```

Inserting:
```sql
INSERT INTO cards VALUES (3, 1, '{"name": "Cook lunch", "tags": ["Cook", "Kitchen", "Tacos"], "ingredients": ["Tortillas", "Guacamole"], "finished": false}');
```

Ability to query JSON:
```sql
SELECT * FROM cards WHERE data->>'finished' = 'true';
```

Creating indexes on JSON attributes:
```sql
CREATE INDEX idxfinished ON cards ((data->>'finished'));
```

WRT question k/v store vs SQL, it's definitely longer discussion.

Gopkg.lock
seanhuxy: will remove

Gopkg.lock
seanhuxy: @kars7e to be honest, I haven't do too much research on different technologies regards to their performance, I will do that and we should discuss about it together tomorrow.

berndtj: Yeah, I would certainly store the values in JSONB columns rather than just text!  I think it's a reasonable step forward without getting into complex schemas and migrations.

Gopkg.lock
seanhuxy: @kars7e 's opinion sounds good to me, but we may (1) implement another identical libkv in our own repo OR (2) lose the flexibility of switching backend db. I understand my version of libkv is not perfect (has bugs), but I suggest that we should keep it and make things work and simple, extensible first. I am also ok with and willing to move it to a vmware controlled repo (a vmware forked libkv)

charts/dispatch/charts/event-manager/templates/deployment.yaml
seanhuxy: Will try this.

pkg/dispatchcli/cmd/install.go
seanhuxy: it is actually, I would also want to have a default config yaml and have the merge function back, WIP on this.

Gopkg.lock
berndtj: Well, I have a feeling in the medium term we may migrate from libkv.  Certainly if we use postgres, there will likely come a time where we actually want to use some of it's features.  I think @kars7e is right in that we already have a defined interface (entity store).  We can code to that, and in the near term we still have all of the advantages.  I'm kind of siding with @kars7e.

config.template.yaml
seanhuxy: I keep my opinion on this, indeed I introduce more values into the config.yaml, but I think most of them should be configurable, although we use the default values all the time, like the namespace, release and etc of each services.

We should have a default config.yaml, and I believe with that you may think this is more clean, small and organized. (the default yaml might be large)

pkg/dispatchcli/cmd/uninstall.go
seanhuxy: 👍  I am working on to have a default config.yaml

pkg/dispatchcli/cmd/uninstall.go
seanhuxy: good catch.

config.template.yaml
berndtj: We just spoke in person, but I think it's good to expose all of the configurable options in one place.  This should be the "default" config so the user only has to override the options they want.  Also, this should be built into the dispatch CLI, so that users don't need download this template.

seanhuxy:
thank @kars7e and @berndtj for very helpful suggestion, I would like to summarize your comments and propose my plans on non-trivial problems.

### libkv or not

considering the requirement/user case of the entity store: there're basically three type of DB Operations
- (1) Operations directly on Primary Key 
  -  CRUD operations by entity Key/ID, i.e. primary key in SQL DB
- (2) Operations on BaseEntity attributes
   - list entities of specific Status, it is heavily used by controllers. ( for best performance, an index is desired and the search should be done with SQL where clause)
   - list entities based on entity types, tags, organizations, and search entities by names (some basic features I expect we will have)
- (3) Operations on Specific Entity attributes
  - e.g. get all APIs pointing to a specific functions.

As we can see, there're some requirements on queries of non-primary key attributes, e.g. (2) (3)

If using libkv abstraction, apparently, the performance of use case (2)(3) will be very low (linear select and filter on Go language level), I guess it will not afford a mass prod deployment.
Based on this, as @kars7e proposed, **we should consider get ride of the libkv layer and build our own DB interface to back the entity store. The DB interface should provide the ability to accept entity store filters and then apply that on a SQL Where statement.**

Based on that, the idea is map the BaseEntity directly into a SQL DB schema/table and build indices on interesting attributes, and use JSONB to store custom entity attributes. JSONB has good query performance, so it fulfills the requirement (3). 

The proposed base entity object looks like this,
```
type BaseEntity struct {
	ID             string    `json:"id"`
	Name           string    `json:"name"`
	OrganizationID string    `json:"organizationId"`
	CreatedTime    time.Time `json:"createdTime,omitempty"`
	ModifiedTime   time.Time `json:"modifiedTime,omitempty"`
	Revision       uint64    `json:"revision"`
	Version        uint64    `json:"version"`
	Spec           Spec      `json:"state"`
	Status         Status    `json:"status"`
	Reason         []string  `json:"reason"`
	Tags           Tags      `json:"tags"`
        Attributes  interface{} `postgres:"JSONB"`  // any custom attributes of a specific entity, still WIP on details.
}
```

### Pure SQL (pq) VS. ORM (go-pg)

the advantage of using ORM is obviously (1) fast and easy to implement, (2) flexible for code/schema change, especially if we decide to abandon KV and embrace SQL for BaseEntity, but I need research on if and how the ORM driver support JSONB, and the corresponding performance.

pure SQL might have better performance, but we should be clear on the question that to how much degree it is better than a ORM solution.

If our plan is to take the BaseEntity as a Relational DB table, I will prefer an ORM solution if that ORM driver also support JSONB

Please feel free to provide any ideas.

imikushin:
I'd suggest using something like https://github.com/jmoiron/sqlx instead of fully fledged ORM

seanhuxy:
Thank you guys for you helpful comments, I made improvement for entity store:
- keep the original libkv + boltdb implementation unchanged
- add an entitystore implementation backed by postgres
- use postgres and JSONB as suggested by @kars7e @berndtj 
- uses pq + sqlx + pure sql as suggested by @kars7e @imikushin @neosab 
- any comments are very welcome, most real changes happen at /pkg/entity-store/postgres.go 

Next Step:
I am going to address comments in the db installation detail and install.go 
then, refine the postgres implementation:
- rename entityStore to something like LibkvEntityStore
- refactor the entity store filter function, so that it can be used in a sql statement
- add an example of how to add index in a JSONB field, which can truly show the advantage of postgres comparing to pure kv store like libkv.

pkg/entity-store/postgres.go
```diff
@@ -0,0 +1,268 @@
...
+	Key    string         `db:"key"`
+	Value  types.JSONText `db:"value"`
+	Status string         `db:"status"`
+}
+
+func (p *postgres) createTable() error {
+
+	sql := `
+		CREATE TABLE IF NOT EXISTS entity (
+		key 		TEXT PRIMARY KEY,
```
berndtj: So we are just dumping everything into the JSONB value column?  Because this is SQL, I would at least assume that ID is now the primary key and that name is a field with an index on it.  Based on the discussion earlier, I thought that most/all of the "Base entity" fields would be columns.

pkg/entity-store/postgres.go
```diff
@@ -0,0 +1,268 @@
...
+	log.Debugf("postgresql database options: %s", opts)
+	db, err := sqlx.Connect("postgres", opts)
+	if err != nil {
+		log.Debugf("error connecting to postgresql DB")
+		return nil, errors.Wrap(err, "Unable to connect to the postgres db server")
+	}
+	store := &postgres{db: db}
+
+	// create tables
+	err = store.dropTable()
```
berndtj: Are we dropping the table every time we start dispatch?  Shouldn't we persist?

pkg/entity-store/postgres.go
berndtj: Probably worth adding a version field as well to deal with future migrations.

pkg/entity-store/postgres.go
```diff
@@ -0,0 +1,268 @@
...
+			return errors.Wrap(err, "error listing entity from db")
+		}
+
+		obj := reflect.New(elemType.Elem())
+		entity := obj.Interface().(Entity)
+		err = row.Value.Unmarshal(entity)
+		if err != nil {
+			return errors.Wrap(err, "error unmarshalling json from db")
+		}
+		// TODO: put filter into sql
```
berndtj: This will get a little tricky with some fields in JSONB and others as columns, but shouldn't be too bad.

pkg/entity-store/postgres.go
seanhuxy: two sides:
first, we could do that, i.e. put shared attributes/fields into as a DB columns, please take the Status field as an example, which requires manually conversion. We can put other fields as a DB columns as we see necessary.

second, ID can the the primary key, but this change will break the the compatibility with the libkv, and require more work on the manager side, if we do this, two options, refactor and use ID for libkv, or, abandon the libkv implementation.

berndtj:
Current changes look pretty clean.  I want to make sure we are settled on the base table schema though.  Maybe others need to chime in?

pkg/entity-store/postgres.go
seanhuxy: good catch, I will remove it

pkg/entity-store/postgres.go
seanhuxy: yea, but I do believe this is the point why we adapt postgres, (the other reason is persistency)

pkg/entity-store/postgres.go
kars7e: I think this is the implementation maps to libkv, but I if we go all-in with postgres, then it would make sense to add normal columns. JSONB still makes sense for something like `properties` - there is almost always a need to add custom, optional fields per entity (vide labels in k8s, metadata in openstack, etc.) those could be implemented via JSONB. 

However, with our current EntityStore interface (specifically, the List function) I don't think we can take advantage of separate columns (faster filtering and indexes).

charts/dispatch/charts/event-manager/templates/deployment.yaml
seanhuxy: @kars7e @berndtj 

this is indeed more concise, but after review the code, I think an aggregated url may not really flexible in terms of compatible to multiple backend. As we discuss before, we want backward compatible with pervious implementation, (in this case an aggregated url does not work with libkv) As long as we haven't decide to we will only use postgres, we may want to keep the flexibility. 

Note: we of course could pass them as a url string and parse it back, but I honestly think it is lose-lose (more work and more error-prone)  

Take Kong as an example, they also have tons of configuration variables, too, and both kong chart and postgres use separate flags/environment variables for values like username, password, and hostnames variables.

PS. although I am not sure which is the best way, but I see many examples from the stable chart repos that they use secret and environment variables to pass sensitive/configuration variables into containerized service, we should think about that.

pkg/entity-store/postgres.go
seanhuxy: indeed given the current filter (it is a function) we can do nothing of indexing or filtering with sql, I consider to refactor the filter to a map-liked structure, so that we could create dynamic sql queries.

for example:
```
filter := map[string]string{
"status":"ready"
}
```
generate a sql like this 
```
select * from entity where status = 'ready'
```
along with an index on columns or within a JSONB field, (for example https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/)

I am optimistic to that. especially most of our use case that requires a filter are equal-based comparison .  

PS. JSONB field will also be used for the custom fields of a specific entity (e.g. API, Image, ...) that the BaseEntity don't contain.

pkg/entity-store/postgres.go
```diff
@@ -0,0 +1,438 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error getting entity from db")
+	}
+
+	return dbToEntity(row, entity)
+}
+
+// List fetches a list of entities of a single data type satisfying the filter.
+// entities is a placeholder for results and must be a pointer to an empty slice of the desired entity type.
+func (p *postgres) List(organizationID string, filter Filter, entities interface{}) error {
```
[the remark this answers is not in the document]
seanhuxy: good points

pkg/entity-store/store.go
```diff
@@ -136,43 +151,117 @@ func (e *BaseEntity) setModifiedTime(modifiedTime time.Time) {
...
 	kv store.Store
 }
 
-// Filter is a function type that operates on returned list results
-type Filter func(Entity) bool
+// Filter defines a set of criteria to filter entities when listing
+type Filter []FilterStat
+
+// FilterStat defines one filter criterion
+type FilterStat struct {
```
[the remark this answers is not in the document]
seanhuxy: Filter Statement, I use it as a single filter statement, will leave a comment there

pkg/entity-store/store.go
```diff
@@ -136,43 +151,117 @@ func (e *BaseEntity) setModifiedTime(modifiedTime time.Time) {
...
+	FilterVerbIn = "in"
+
+	// FilterVerbEqual tests equality
+	FilterVerbEqual = "equal"
+
+	// FilterVerbBefore tests two time.Time
+	FilterVerbBefore = "before"
+
+	// FilterVerbAfter tests two time.Time
+	FilterVerbAfter = "after"
```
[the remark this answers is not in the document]
seanhuxy: good point

seanhuxy:
run e2e locally and passed. The postgres entity store should now work with all unit test and e2e test.

however, it have potential problems if there are new/complicated queries to entity store. (it is mainly because,, instead of writing a customized filter function, now we generate filter logic from a set of filter verbs, which of course has limitations.)

I want to list unsupported cases here, but I do not think it will be needed or we should make it such perfect now. With this PR, we have a basic framework and the extensibility to add new/complex filter verbs as we see new needs.
- cannot filter on field tags, spec and reason, as they are jsonb type, it requires more work to be able to filter on them. Potentially we may need filter on tags, (but we don't have use case right now).
- only able to filter simple type field of a custom entity, e.g. getting run by function names. not able to filter on e.g. getting runs which logs contains an specific words (again, I don't actually expect such a complicated query, so it should be fine)
- filters are AND'ed, complicated filter like ``getting runs created in 2017 OR created by user Bernd`` is not supported, which I also don't find use cases.

In a word, for current use cases, it should work well.

charts/dispatch/charts/event-manager/templates/deployment.yaml
berndtj: This isn't that big a deal... Though even if the back end takes a list of parameters and not a URL, we could always parse the URL.  Not a big deal though

charts/dispatch/values.yaml
```diff
@@ -15,5 +15,15 @@ global:
...
   tls:
     secretName: dispatch-tls
+  db:
+    backend: postgres
+    host: postgresql
+    port: 5432
+    user: vmware
+    password: dispatch
+    database: dispatch
+    # namespace: <set if different from dispatch>
```
berndtj: If these are settable values, it's nice to actually put the default in here.  Otherwise remove.

pkg/entity-store/store.go
```diff
@@ -299,6 +430,62 @@ func (es *entityStore) Get(organizationID string, name string, entity Entity) er
 	return nil
 }
 
+func doFilterStat(fs FilterStat, entity Entity) (bool, error) {
```
berndtj: So are we not using SQL to do the actual filtering in the case of Postgres?

pkg/entity-store/postgres.go
```diff
@@ -0,0 +1,439 @@
...
+		entity := entityPtr.Interface().(Entity)
+
+		dbToEntity(row, entity)
+		slice = reflect.Append(slice, entityPtr)
+	}
+	rv.Elem().Set(slice)
+	return nil
+}
+
+// Delete delets a single entity from the store.
```
berndtj: nit s/delets/deletes/

pkg/entity-store/store.go
```diff
@@ -136,43 +151,117 @@ func (e *BaseEntity) setModifiedTime(modifiedTime time.Time) {
...
+func (e *BaseEntity) GetSpec() Spec {
+	return e.Spec
+}
+
+// GetTags retreives the entity tags
+func (e *BaseEntity) GetTags() Tags {
+	return e.Tags
+}
+
 type entityStore struct {
```
berndtj: I think I was a bit confused because we should probably rename this to KVEntityStore or something more descriptive.

pkg/entity-store/store.go
seanhuxy: nope, this function is for libkv filter, please see the function MakeListSQL function in postgres.go

pkg/entity-store/store.go
seanhuxy: Yeah, I want to do that, too.

pkg/entity-store/postgres.go
seanhuxy: thanks

charts/dispatch/values.yaml
seanhuxy: it's some leftovers, but good catches, I will fix it.

PR #83, opened 2018-01-02
WIP: Add riff FaaS driver

imikushin:
Basic support for [riff](https://projectriff.io)


pkg/functions/riff/builder.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (d *riffDriver) processRequests() {
+	for r := range d.buildRequests {
+		r.result <- d.buildAndPushImage(r)
+		close(r.result)
+	}
+}
+
+func dockerError(r io.ReadCloser, err error) error {
```
berndtj: This function is copied from openfaas/builder.go.  Can we just make it common?

pkg/functions/riff/builder_test.go
```diff
@@ -0,0 +1,27 @@
+///////////////////////////////////////////////////////////////////////
+// Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+///////////////////////////////////////////////////////////////////////
+package riff
```
berndtj: Could we have a "common" test suite against the different drivers and builders?  Obviously they would have external dependencies, but it may be a good integration/e2e type test.

pkg/functions/riff/builder.go
imikushin: yeah, making the builder a common util is on my list

pkg/functions/riff/builder_test.go
imikushin: I'll think about it. I try to test things bottom up. But we sure need e2e tests for these.

PR #87, opened 2018-01-04
Set correct dispatch port in .dispatch/config.json

### LINKED ISSUE #82: Dispatch install always puts port 443 in CLI config
After installation is complete, dispatch puts port `443` in `.dispatch/config.json`, regardless of the fact how ingress is installed (whether LoadBalancer or NodePort is used). Since dispatch takes care of ingress installation, it should detect and use correct port when ingress is installed with non-standard port.

neosab:
Currently, `dispatch install` always generates the dispatch
`~/.dispatch/config.json` with port `443`. When using NodePort
as the ServiceType, a high port is assigned for the ingress
controller service and that must be used as the port of
the dispatch service.

* This patch fixes the config json by fetching the assigned NodePort.
* The patch removes the hostPort option in the container
  spec for the ingress deployment. Kubernetes doesn't recommend
  using hostPort as it creates scheduling constraints and the port
  may not be available on hosted K8S environments (it's possible
  the apiserver might be using it).
* The patch fixes the redirect_uri of the oauth2proxy configmap
   so that the auth request can be redirected to the appropriate NodePort.
  
Fixes #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,34 +74,41 @@ oauth2Proxy:
 EOF
 ```
 
-Install and copy the configuration to `$HOME/.dispatch.json`:
+Run Install:
 ```
 $ ./dispatch-darwin install --file config.yaml --service all --debug
 ...
-Copy the following to your $HOME/.dispatch.json
+Config file written to: $HOME/.dispatch/config.json
+```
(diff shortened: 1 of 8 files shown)


neosab:
@berndtj Done!

PR #89, opened 2018-01-04
Do not push images during the build by default

imikushin:
We don't want to push stuff to devs' DockerHub accounts w/o consent.

Use `PUSH_IMAGES=1 make images` to push.


scripts/images.sh
```diff
@@ -14,6 +14,6 @@ image=${DOCKER_REGISTRY}/dispatch-${PACKAGE}:${TAG}
 echo $image
 
 docker build -t $image -f images/${PACKAGE}/Dockerfile .
-if [ -z $NO_PUSH ]; then
+if [ -z $NO_PUSH ] && [ -n "$PUSH_IMAGES" ]; then
```
kars7e: Seems like NO_PUSH is not needed anymore?

imikushin: I'm gonna remove it, if it's not used by anyone.

kars7e: I think it's safe to remove, because with your change this variable has no effect anymore

PR #96, opened 2018-01-09
Support dispatch installation with User-Provided Certs

neosab:
Currently, we generate the required certs for tls in `dispatch install`. We do this for both the dispatch ingress endpoint and the api-gateway.

This patch adds support for securing the endpoints with user-provided certs.

The README will be enhanced later since we are missing instructions to customize the dispatch install's `config.yaml` and the file format is in flux.

### DIFF
--- a/pkg/dispatchcli/cmd/install.go
+++ b/pkg/dispatchcli/cmd/install.go
@@ -54,12 +54,19 @@ type postgresConfig struct {
 	Persistence bool         `json:"persistence,omitempty" validate:"omitempty"`
 }
 
+type tlsConfig struct {
+	CertFile   string `json:"certFile,omitempty"`
+	PrivateKey string `json:"privateKey,omitempty"`
+	CertSecret string `json:"certSecret,omitempty" validate:"required"`
+}
+
 type apiGatewayConfig struct {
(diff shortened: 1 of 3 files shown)


pkg/dispatchcli/cmd/install.go
```diff
@@ -218,22 +240,33 @@ func NewCmdInstall(out io.Writer, errOut io.Writer) *cobra.Command {
...
-		if _, err = os.Stat(cert); os.IsNotExist(err) {
-			openssl := exec.Command(
-				"openssl", "req", "-x509", "-nodes", "-days", "365", "-newkey", "rsa:2048",
-				"-keyout", key,
-				"-out", cert,
-				"-subj", subject)
-			opensslOut, err := openssl.CombinedOutput()
-			if err != nil {
-				return errors.Wrapf(err, string(opensslOut))
+func installCert(out, errOut io.Writer, configDir, namespace, domain, certName string, tls *tlsConfig) error {
```
berndtj: So do we no longer support making SSL certs on behalf of the user?  I'd like to keep that option for super easy single step setup.

neosab:
@kars7e We will update the README soon as there isn't yet a section to discuss advanced configurations during installation. I don't want to add it to QuickStart.

PR #99, opened 2018-01-09
Adding back all existing documentation

berndtj:
* using Jekyll to post documentation to github pages
* specs are known to be out of date, but we will address that
* navigation is flat currently, but we will address that


kars7e:
How will one access the website generated by Jekyll in this case? Could we add CNAME to support https://dispatchframework.io, or it won't work with subdirectory?

berndtj:
@kars7e You can access the docs at vmware.github.io/dispatch.  I believe you can also add CNAME support, but I'll have to look into it.

PR #105, opened 2018-01-10
[cli/update] Add Update functionality for secrets

tenczar:


pkg/dispatchcli/cmd/update.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/spf13/cobra"
+	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
+)
+
+var (
+	updateLong = i18n.T(`Create a resource. See subcommands for resources that can be created.`)
+
+	// TODO: Add examples
+	updateExample = i18n.T(``)
+	//file          = i18n.T(``)
```
berndtj: delete this

pkg/dispatchcli/cmd/update.go
```diff
@@ -0,0 +1,46 @@
...
+			if file == "" {
+				runHelp(cmd, args)
+				return
+			}
+
+			updateMap := map[string]modelAction{
+				"secret": CallUpdateSecret,
+			}
+
+			err := importFile(out, errOut, cmd, args, updateMap)
```
berndtj: Does this actually work?  It doesn't seem as though you've added a file argument to the update command.

pkg/secret-store/service/k8s_secrets_service.go
```diff
@@ -131,20 +132,28 @@ func (secretsService *K8sSecretsService) DeleteSecret(name string) error {
 }
 
 func (secretsService *K8sSecretsService) UpdateSecret(secret models.Secret) (*models.Secret, error) {
-	var entities []*secretstore.SecretEntity
+	entity := secretstore.SecretEntity{}
+	name := *secret.Name
 
+<<<<<<< HEAD
```
berndtj: Fix the merge

tenczar: oh... It looks like my IDE does not automatically save after I tell it to take one change over another... I can fix this quickly.

pkg/dispatchcli/cmd/update.go
tenczar: I admit that I took this from the code that sets up the create command. Now that I look at this, I'm just as confused as you are. It compiles and runs properly, but I can't tell where that `file` variable comes from.

tenczar: Actually I have an idea... I believe file is declared as a package scoped variable in the create.go file. In that file it is declared as `file = i18n.T(\`\`)`

PR #108, opened 2018-01-10
update config.yaml in README.md

seanhuxy:
in addition to this, the dispatch images should be updated in the dispatch image repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,22 +55,24 @@ You should have ``<oauth-client-id>``, ``<oauth-client-secret>`` and ``<oauth-co
 Configure the installation.  Substitute in your docker credentials (host and username are likely the same):
 ```
 $ cat << EOF > config.yaml
-namespace: dispatch
-hostname: dev.dispatch.vmware.com
-certificateDirectory: /tmp
-chart:
+apiGateway:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,22 +55,24 @@ You should have ``<oauth-client-id>``, ``<oauth-client-secret>`` and ``<oauth-co
...
-  host: <docker repo>
-  username: <docker username>
-  email: <docker email>
-  password: <docker password>
-oauth2Proxy:
-  clientID: <oauth2-client-id>
-  clientSecret: <oauth2-client-secret>
-  cookieSecret: <oauth2-cookie-secret>
+  debug: true
+  openfaasRepository:
```
berndtj: I realize this isn't about this checkin, but this won't *just* be an openFaaS repo.  This is a dispatch repo used to store images as well as any function artifacts.

seanhuxy: hhh, I used "openfaasRepository" when I thought it was only used by openfaas to store function images/artifacts. And yeah, probably dispatchRepository or imageRepository might be better. (especially maybe in the future when this repository is not only used openfaas)

PR #110, opened 2018-01-10
Fixup small issues with regards to deploying

berndtj:
* Fix any inconsistencies in README.md
* Add extra debug output
* Add roadmap items
* Bump chart version

### DIFF
--- a/pkg/dispatchcli/cmd/install_config.go
+++ b/pkg/dispatchcli/cmd/install_config.go
@@ -52,7 +52,7 @@ dispatch:
   certSecret: dispatch-tls
   image:
     host: vmware
-    tag: dev-xxxx
+    tag: v0.1.2
   database: postgres
   debug: true
   trace: true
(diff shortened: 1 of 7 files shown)


pkg/dispatchcli/cmd/install_config.go
```diff
@@ -52,7 +52,7 @@ dispatch:
   certSecret: dispatch-tls
   image:
     host: vmware
-    tag: dev-xxxx
+    tag: v0.1.2
```
neosab: version bump to v0.1.2  should be done on [L46](https://github.com/vmware/dispatch/pull/110/files#diff-04c6e90faac2675aa89e2176d2eec7d8R46) and [L65](https://github.com/vmware/dispatch/pull/110/files#diff-04c6e90faac2675aa89e2176d2eec7d8R46) as well.

berndtj: thanks!

PR #114, opened 2018-01-11
Example: Buzzword Bingo slack bot

imikushin:
A real Slack bot implemented as a function: posts "BINGO!" when someone says
things like "cloud native serverless platform" (looks for 3 or more 
buzzwords in your posts) on Slack.

### DIFF
--- a/examples/bingo/base-image/Dockerfile
+++ b/examples/bingo/base-image/Dockerfile
@@ -1,0 +1,3 @@
+FROM vmware/dispatch-openfaas-nodejs6-base:0.0.3-dev1
+
+RUN npm install request@2.83.0
--- a/examples/bingo/bingo.js
+++ b/examples/bingo/bingo.js
@@ -1,0 +1,43 @@
+"use strict";
+
+const request = require('request');
+
+const buzzwords = ["cloud", "native", "serverless", "platform", "infrastructure", "integration", "framework"];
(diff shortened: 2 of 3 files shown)


examples/bingo/bingo.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
+
+const request = require('request');
+
+const buzzwords = ["cloud", "native", "serverless", "platform", "infrastructure", "integration", "framework"];
+
+function bsCount(text) {
```
kars7e: lol

examples/bingo/base-image/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM vmware/dispatch-openfaas-nodejs6-base:0.0.3-dev1
+
+RUN npm install request@2.83.0
```
kars7e: I actually need similar image for https://github.com/vmware/dispatch/blob/master/examples/nodejs6/slack_vm_being_deployed.js 
Should we create few more "common" images in `vmware` org (or some separate org, like dispatch) until we add the feature to build images based on requirements.txt to dispatch? CC @berndtj

examples/bingo/bingo.js
imikushin: 😸

Issue #80, opened 2017-12-23
Clarify documentation on docker credentials and FQDN


markpeek:
The documentation should clarify:
1. docker credentials
  - What will the docker credentials be used for?
  - Clarify the difference between host and username.
2. FQDN
  - A FQDN is required to be put into /etc/hosts, what is is used for?
  - The documentation should be using example.com per [RFC6761](https://tools.ietf.org/html/rfc6761)

PR #125, opened 2018-01-12
Make Image Registry host optional and some README changes

neosab:
If you are using dockerhub, the image registry host must be specified the same as your username. This is kind of confusing to the end user. We can make it optional and default it to the username
if not specified (i.e assume docker hub).

Some other changes include changing the domain to example.com and
formatting the README's QuickStart.

Partially addresses #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,14 @@ The diagram below illustrates the different components which make up the Dispatc
 
 ## Quickstart
 
-Assumptions:
+### Assumptions:
 
 * Minikube is installed.  If not, simply download and install minikube
   (See [minikube](#minikube)).
 * Kubectl installed and configured for minikube.
 * Running on MacOS
 
+### Start Minikube and initialize Helm
 ```
 # Start minikube (if not already started)
 $ minikube start --vm-driver=hyperkit --bootstrapper=kubeadm --disk-size=50g --memory=6144 --kubernetes-version=v1.8.1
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -38,45 +39,54 @@ $ helm init
...
-  hostname: api.dev.dispatch.vmware.com
+  hostname: api.dev.dispatch.example.com
 dispatch:
-  hostname: dev.dispatch.vmware.com
+  hostname: dev.dispatch.example.com
   image:
     host: vmware
     tag: v0.1.2
   debug: true
   openfaasRepository:
```
kars7e: can you also change this? It should be registry, not repository, and it should not be tied to openfaas - it should probably be just named `imageRegistry`. It would require a change in CLI as well, so feel free to ignore if you think it should go into separate PR (but I think we should address this quickly as this is a breaking change).

neosab: Yeah, I thought will make it separately but yeah could do here as well not a big deal. Will push another change.

pkg/dispatchcli/cmd/install.go
```diff
@@ -80,10 +81,10 @@ type imageConfig struct {
 type oauth2ProxyConfig struct {
 	ClientID     string `json:"clientID,omitempty" validate:"required"`
 	ClientSecret string `json:"clientSecret,omitempty" validate:"required"`
-	CookieSecret string `json:"cookieSecret,omitmepty" validate:"omitempty"`
+	CookieSecret string `json:"cookieSecret,omitempty" validate:"omitempty"`
 }
 type openFaasRepoConfig struct {
-	Host     string `json:"host,omitempty" validate:"required,hostname"`
+	Host     string `json:"host,omitempty" validate:"omitempty"`
```
berndtj: So, I'm not sure we want to default to docker hub.  In fact, sooner (hopefully very soon), we should deploy a registry with dispatch as a default (Harbor).  Optionally, you can point to an existing registry such as docker hub or artifcatory.

README.md
berndtj: 👍

pkg/dispatchcli/cmd/install.go
neosab: We could then make it explicit. Like, specify the registry as 'index.docker.io' in order to use dockerhub?  I just want the 'host' configuration to be either an FQDN or nothing. Substituting a username in a host field is what I am trying to avoid. Thoughts?

berndtj: I'm ok with being explicit.  I think most of this really comes down to documentation.

neosab: Let's then keep it required for now and explicitly specify index.docker.io to use dockerhub and when we have our own dispatch registry we can make this optional.

neosab:
Some of the changes done in the original patch like renaming to example.com were merged in patches from others. Closing this PR & fully addressing issue #80 in another PR #147.

Issue #41, opened 2017-12-13
Control-loop pattern for image-manager


kars7e:
Image manager already uses a version of the control-loop, but use shared libraries to follow patterns of function-manager and api-manager.

PR #128, opened 2018-01-17
Integrate control loop library with image manager

berndtj:
* Add sync method to EntityHandler interface
  - used to fetch work and is configurable for each entity type
  - image manager compares current state with stored state and
    returns list of entities which need work
  - DefaultSync provides existing behavior to services
* Make work queue scale up as opposed to just starting n (1000)
  goroutines by default

Testing Done: ran e2e
Fixes: #41

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -6,11 +6,13 @@
 package controller
 
 import (
+	"context"
 	"reflect"
 	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 
 	"github.com/vmware/dispatch/pkg/entity-store"
 	"github.com/vmware/dispatch/pkg/trace"
@@ -23,6 +25,9 @@ type EntityHandler interface {
 	Update(obj entitystore.Entity) error
 	Delete(obj entitystore.Entity) error
 	Error(obj entitystore.Entity) error
+	// Sync returns a list of entities which to process.  This method should call out and determine the actual state
+	// of entities.
+	Sync(organizationID string, resyncPeriod time.Duration) ([]entitystore.Entity, error)
 }
 
 const defaultWorkers = 1
@@ -36,8 +41,10 @@ type Options struct {
 	Workers      int
 }
 
+// Watcher channel type
 type Watcher chan<- entitystore.Entity
 
+// OnAction pushes an entity onto the watcher channel
 func (w *Watcher) OnAction(e entitystore.Entity) {
 	defer trace.Trace("")()
 
@@ -62,14 +69,13 @@ type Controller interface {
 type DefaultController struct {
 	done    chan bool
 	watcher chan entitystore.Entity
-	store   entitystore.EntityStore
 	options Options
 
 	entityHandlers map[reflect.Type]EntityHandler
 }
 
 // NewController creates a new controller
-func NewController(store entitystore.EntityStore, options Options) Controller {
+func NewController(options Options) Controller {
 	defer trace.Trace("")()
 
 	if options.Workers == 0 {
@@ -79,7 +85,6 @@ func NewController(store entitystore.EntityStore, options Options) Controller {
 	return &DefaultController{
 		done:    make(chan bool),
 		watcher: make(chan entitystore.Entity),
-		store:   store,
 		options: options,
 
 		entityHandlers: map[reflect.Type]EntityHandler{},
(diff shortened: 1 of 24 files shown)


pkg/image-manager/image_builder.go
```diff
@@ -145,92 +129,41 @@ func (b *BaseImageBuilder) dockerStatus() ([]*BaseImage, error) {
...
-				err = b.dockerPull(&bi)
+		} else {
+			// If the image is present, move to READY if in a
+			// non-DELETING statue
+			switch s := bi.Status; s {
+			case entitystore.StatusINITIALIZED,
+				entitystore.StatusCREATING,
+				entitystore.StatusUPDATING,
+				entitystore.StatusERROR:
+				bi.Status = entitystore.StatusREADY
```
kars7e: ugh, formatting on this one is weird. Is this gofmt output? It's hard to see where do the case options end and its code starts.

berndtj: I know... it's gofmt.  I had it more clear, but it just got moved back :(

pkg/controller/controller.go
```diff
@@ -79,7 +86,7 @@ func NewController(store entitystore.EntityStore, options Options) Controller {
 	return &DefaultController{
 		done:    make(chan bool),
 		watcher: make(chan entitystore.Entity),
-		store:   store,
+		// store:   store,
```
kars7e: NIT: remove the commented line

pkg/controller/controller.go
```diff
@@ -62,14 +69,14 @@ type Controller interface {
 type DefaultController struct {
 	done    chan bool
 	watcher chan entitystore.Entity
-	store   entitystore.EntityStore
+	// store   entitystore.EntityStore
```
seanhuxy: remove

pkg/controller/controller.go
seanhuxy: remove

pkg/controller/controller.go
```diff
@@ -121,22 +128,27 @@ func (dc *DefaultController) processItem(e entitystore.Entity) error {
 	if !ok {
 		return errors.Errorf("trying to process an entity with no entity handler: %v", reflect.TypeOf(e))
 	}
+	if e.GetDelete() {
```
seanhuxy: I think the Delete is a legacy flag that should be cleaned, it should be replaced by the status ``StatusDELETING ``

pkg/controller/controller.go
```diff
@@ -186,21 +211,27 @@ func (dc *DefaultController) run(stopChan <-chan bool) {
...
-
-			for entity := range dc.watcher {
-				func() {
-					log.Printf("received event=%s entity=%s", entity.GetStatus(), entity.GetName())
-					if err := dc.processItem(entity); err != nil {
-						log.Error(err)
-					}
-				}()
+	// Start a worker pool.  The pool scales up to dc.options.Workers.
+	go func() {
```
kars7e: this only spawns one goroutine, how does it scale to `dc.options.Workers`?

berndtj: Doh!  forgot one little key word

kars7e: Magic of Go, instead of few imports and multiple declarations, now you only need to miss two characters to make your application accidentally single-threaded ;)

pkg/image-manager/handlers.go
```diff
@@ -370,9 +369,9 @@ func (h *Handlers) deleteImageByName(params image.DeleteImageByNameParams, princ
 	}
 	e.Delete = true
```
seanhuxy: remove this flag, I think it is duplicated with e.Status

pkg/image-manager/image_builder.go
```diff
@@ -122,16 +111,11 @@ func (b *BaseImageBuilder) dockerDelete(baseImage *BaseImage) error {
...
-	err = b.es.Delete(b.orgID, baseImage.Name, &deleted)
-	if err != nil {
-		return errors.Wrapf(err, "Error deleting base image entity %s/%s", baseImage.OrganizationID, baseImage.Name)
-	}
 	log.Printf("Successfully deleted base image %s/%s", baseImage.OrganizationID, baseImage.Name)
 	return nil
 }
 
-func (b *BaseImageBuilder) dockerStatus() ([]*BaseImage, error) {
+func (b *BaseImageBuilder) dockerStatus() ([]entitystore.Entity, error) {
```
seanhuxy: I think it is better to be named as baseImageStatus, baseImageDelete and baseImagePull for BaseImageBuilder if you used imageStatus, imageDelete and imagePull for ImageBuilder

pkg/controller/controller.go
```diff
@@ -157,17 +169,30 @@ func toSync(resyncPeriod time.Duration) entitystore.Filter {
...
+}
+
 func (dc *DefaultController) sync() error {
 	defer trace.Trace("")()
 
-	for entityType := range dc.entityHandlers {
-		entitiesPtr := reflect.New(reflect.SliceOf(entityType))
-		if err := dc.store.List(dc.options.OrganizationID, toSync(dc.options.ResyncPeriod), entitiesPtr.Interface()); err != nil {
+	for _, handler := range dc.entityHandlers {
+		entities, err := handler.Sync(dc.options.OrganizationID, dc.options.ResyncPeriod)
```
seanhuxy: I think this is a very good refactor. 

As the real intention of a Sync function is to find the difference between the desired status (in entity store) and the actual status (in underlying driver); then try to resolve the difference. 

I gave up and decide to work on the easier implementation first, i.e. just enforce the actual status towards the desired status. In which case, there may be some zombie entities/resources in the underlying drivers.  

I think the new Sync function of Image Manager did a good job on this, (i.e. also pull image status in the driver (image registry). Hopefully, we can apply a similar pattern for other entities. (however, sometimes it may be error-prone, difficult or inefficient to periodically "deeply" compare two set of objects/status

pkg/controller/controller.go
berndtj: I'm not sure about this.  I don't really like overloading the status field with the "requested state" which is what we are doing here.  The current state when deleting something is still "READY", but we are flagging it for deletion.  This is going to be a bigger conversation at some point, where we may need to more formally use requested state (currently the spec field on the entity) to reflect the desired state, while status is simply the current state.

pkg/image-manager/handlers.go
berndtj: see earlier comment.

pkg/controller/controller.go
imikushin: 👍 

LOL "zombie entities" 😂

pkg/controller/controller.go
imikushin: I tend to agree with @berndtj on this.

PR #135, opened 2018-01-22
[cli/update] Add tests to secret updates for cli integration

tenczar:
Add unit tests and e2e test for secret update. Also modified update code to remove dependency on libkv.

### DIFF
--- a/pkg/secret-store/service/k8s_secrets_service.go
+++ b/pkg/secret-store/service/k8s_secrets_service.go
@@ -131,26 +131,17 @@ func (secretsService *K8sSecretsService) DeleteSecret(name string) error {
 }
 
 func (secretsService *K8sSecretsService) UpdateSecret(secret models.Secret) (*models.Secret, error) {
-	name := secret.Name
-	nameFilter := []entitystore.FilterStat{
-		entitystore.FilterStat{
-			Subject: "Name", Verb: entitystore.FilterVerbEqual, Object: name,
-		},
-	}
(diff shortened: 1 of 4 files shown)


pkg/secret-store/service/k8s_secrets_service.go
```diff
@@ -132,20 +131,26 @@ func (secretsService *K8sSecretsService) DeleteSecret(name string) error {
...
-		return nil, SecretNotFound{}
+	name := secret.Name
+	nameFilter := []entitystore.FilterStat{
+		entitystore.FilterStat{
+			Subject: "Name", Verb: entitystore.FilterVerbEqual, Object: name,
+		},
 	}
 
+	var entities []*secretstore.SecretEntity
+	err := secretsService.EntityStore.List(secretsService.OrgID, nameFilter, &entities)
```
berndtj: Why are you calling list as opposed to Get?

tenczar: The previous logic relied on checking an error that was only thrown by libkv to determine if the entity was not found. That error is not thrown by postgres. This implementation allows me to determine if the entity was not found without relying on those errors.

berndtj: What does Get return then? The point is, that it sounds like you are highlighting a bug in the entity store.  If this is the case, then we need to fix it, not just work around it by using List everywhere.  This will be an issue everywhere we use Get, so we should just fix it.

PR #139, opened 2018-01-24
application specifications

seanhuxy:
- also update roadmap, and fix broken links in it

### DIFF
--- a/docs/_front/roadmap.md
+++ b/docs/_front/roadmap.md
@@ -49,24 +49,15 @@ metadata should be propogated through the system (associated with the request or
 
 ## 1.3 Applications or Groups
 
-Dispatch requires a grouping mechanism in order to better structure resources such as functions and API endpoints.
-The suggested name for this grouping is "application", though that is subject to change.  Application is itself a
-first class resource, belonging to an organization.  Application should be a required property of the following
-resources:
+Application is one of the most important concept of Dispatch. It is a grouping and isolating mechanism in order to better structure resources such as functions and API endpoints.
 
-* Function
-* API
-* Secret
-* Subscription (may be implied through function)
+In addition to simply group resources, applications provides system admins and developers much more: application-level resource management, permission management and access controls; in the further, application-level lifecycle management, function execution workflows, application-level monitoring and analyzing, etc.
 
-Base-images and images may remain tied to an organization.  It's possible that they may optionally be tied to an
-application.
-
-Applications should additionally enable per-application hostname (domains?) and certificates on the API gateway.
+Please see the [application specification](../specs/application/application.html) for detailed descriptions.
 
 ## 1.4 Image Management
 
-Currently the image support in Dispatch is simply pass-through. See the [image manager spec](image-manager.md) for
+Currently the image support in Dispatch is simply pass-through. See the [image manager spec](../specs/image-manager/image-manager.html) for
 feature description.
 
 ## 1.5 Multi-Organization Support
--- a/docs/_specs/application/application.md
+++ b/docs/_specs/application/application.md
@@ -1,0 +1,89 @@
+---
+layout: default
+---
+
+# Application-scoped Dispatch
+
+## Introduction
+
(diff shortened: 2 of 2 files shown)


docs/_specs/application/application.md
```diff
@@ -0,0 +1,92 @@
...
+
+In addition to those basic functionalities, the concept of Application also provides a step forwards to advanced features of cloud native apps, e.g. lifecycle management (i.e. creation/delete, import and export Applications, scaling up and down based on demand/usage);, function execution workflow management, visualized workflow GUI; application level monitoring and analyzing; function/application external dependencies management.
+
+In this design documents, we outline several design concerns and implementation decisions with regards to the concept of a Serverless Application.
+
+Note: there is another grouping concept “Organization” within the realm of Dispatch. As its detail definition hasn't been finalized, and it is out of the scope of this articles; we decide to only discuss “Application” under the assumption of a single tenant dispatch deployment, in which the system scope and the organization scope are equivalent, and the system admin and the organization admin are equivalent.
```
berndtj: You could probably just remove this note.

docs/_specs/application/application.md
```diff
@@ -0,0 +1,92 @@
...
+Both shared one-to-many and many-to-many events are valid use cases. When there’s not enough real use cases it is hard to decide which is dominant and which is rare. Again, we decide to keep it flexible at this point. In the first iteration, we will set them as private to an Application for now.
+
+As a side note, although Event Topics are not a dispatch entity now, we should also consider to scope them by Application (especially if we decide that Event Subscriptions should be scoped). For example, add the Application name as a prefix of a topic name.
+
+## Implementation Detail
+
+
+### The Application Entity
+
+As most entities are organized by applications, an Application itself also has a life cycle. To manage applications, it is best to manage and maintain it with the existing Entity Store.
```
berndtj: What happens to entities which reference applications when applications are deleted?  How does the application lifecycle impact the entities which reference the application?

seanhuxy: that's a good question, with many-to-many relationship, I can image we should not delete associated entities if one application is deleted. 

we should also probably have a DeleteOption say with DeleteOption = "cascaded", delete all associated entities, with DeleteOption="regular", only the application entity will be delete.

docs/_specs/application/application.md
seanhuxy: sure

docs/_specs/application/application.md
berndtj: I think we should cascade in this scenario.  Also for APIs.  They should probably be cascaded when a function is deleted.

seanhuxy: got you, we could have the default behavior as cascaded for now.

docs/_front/roadmap.md
```diff
@@ -49,24 +49,15 @@ metadata should be propogated through the system (associated with the request or
...
-Base-images and images may remain tied to an organization.  It's possible that they may optionally be tied to an
-application.
-
-Applications should additionally enable per-application hostname (domains?) and certificates on the API gateway.
+Please see the [application specification](../specs/application/application.html) for detailed descriptions.
 
 ## 1.4 Image Management
 
-Currently the image support in Dispatch is simply pass-through. See the [image manager spec](image-manager.md) for
+Currently the image support in Dispatch is simply pass-through. See the [image manager spec](../specs/image-manager/image-manager.html) for
```
imikushin: Broken link

seanhuxy: actually, tested with a local github page server, found the ``image-manager.md`` is broken, ``../specs/image-manager/image-manager.html`` works for me. 

if you see the current ``https://vmware.github.io/dispatch/front/roadmap.html``, it is broken currently.

docs/_specs/application/application.md
kars7e: Kubernetes has an option, enabled by default, to cascade delete the resources when deleting a namespace. We could follow this approach (delete by default).

PR #140, opened 2018-01-24
riff FaaS driver

imikushin:
Official riff-0.0.3-rbac chart is used.

Either `openfaas` or `riff` can be specified in install config.yaml:
`  faas: riff`
Only the configured driver gets installed. OpenFaaS gets installed by
default.

`dispatch uninstall` tolerates not found services.

Function images are named as follows:
`<registry>/func-<faas>-<fnID>`

Function ID are used in lieu of Name in FaaS drivers.

Don’t delete the function in the FaaS driver while creating it.
Simply create. If we’re using UUID there’s no such function in yet.

e2e tests can run with riff driver:
`INSTALL_DISPATCH=0 e2e/scripts/run-e2e.sh ./e2e/tests-riff`

### DIFF
--- a/charts/dispatch/charts/function-manager/values.yaml
+++ b/charts/dispatch/charts/function-manager/values.yaml
@@ -42,6 +42,9 @@ faas:
     host: "52.91.175.16"
   openfaas:
     gateway: "http://gateway.openfaas:8080/"
+  riff:
+    gateway: "http://riff-riff-http-gateway.riff/"
+    namespace: riff
 registry: {}
   # insecure: false
   # uri: docker-docker-registry.docker.svc.cluster.local:5000
--- a/e2e/tests-riff/apis.bats
+++ b/e2e/tests-riff/apis.bats
@@ -1,0 +1,116 @@
+#!/usr/bin/env bats
+
+set -o pipefail
+
+load helpers
+load variables
+
+: ${API_GATEWAY_HTTPS_HOST:="https://api.dev.dispatch.vmware.com:${API_GATEWAY_HTTPS_PORT}"}
+: ${API_GATEWAY_HTTP_HOST:="http://api.dev.dispatch.vmware.com:${API_GATEWAY_HTTP_PORT}"}
+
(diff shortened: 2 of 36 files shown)


berndtj:
Is this review up to date?

charts/dispatch/charts/function-manager/values.yaml
```diff
@@ -42,8 +42,12 @@ faas:
     host: "52.91.175.16"
   openfaas:
     gateway: "http://gateway.openfaas:8080/"
-    imageRegistry: "vmware"
-    registryAuth: SECRET
+  riff:
+    gateway: "http://riff-riff-http-gateway.riff/"
```
berndtj: that's a LOT of riff

e2e/tests-riff/apis.bats
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bats
```
berndtj: There's probably a better way than creating separate riff tests, but we can fix that separately

pkg/entity-store/store.go
```diff
@@ -29,6 +29,11 @@ const (
 	// should wait until the READY state to use the object
 	StatusCREATING Status = "CREATING"
 
+	// StatusINTRANSIT object is currently being processed
+	// and should NOT be picked up by periodic sync
+	// TODO(imikushin) GC stale inflight objects - in case responsible process terminates (e.g. using process UUIDs)
+	StatusINTRANSIT Status = "INTRANSIT"
```
berndtj: We talked about this.  This isn't a good status as it doesn't imply what transition is happening, we lose context.  I'm not going to hold back this PR, but this isn't a pattern I'd like to push through dispatch.

Also, it doesn't convey any meaning to the user.  The actual status should be "CREATING" which means it is in transition from INITIALIZED -> READY.  The function could be created with the INITIALIZED initial state (like images)

imikushin:
Fixes #159

Issue #90, opened 2018-01-04
Add "application" or similar construct to Dispatch


berndtj:
Functions, images, APIs, etc. should be groupable.  We need a new grouping construct "application" (though could use different name as well).  This will have a large impact to just about all services.

- [ ] Bug
- [x] Feature
- [ ] Enhancement

## Detailed Description

In order to support multiple serverless applications or deployments on a single dispatch installation, a grouping construct is required.  Additionally, this construct will be used for authorization and permissions in the future.  Similar to a namespace in k8s, "applications" can be created and managed, then resources get assigned to them.

## Context

This is vital towards making dispatch usable in a production environment

## Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [x] High - Significant architectural change or large refactor

## Impact

- [ ] Low - Annoyance, but does not impact business or functinality
- [ ] Medium - Issue can be worked around, but is causing pain
- [x] High - Blocker

## Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

seanhuxy:
as discussed with @berndtj I am going to work on this feature. Will have a detailed refactor proposal within several days

berndtj:
Split into another issue to capture remaining work

seanhuxy:
first task: add application manager, provide basic functionality of an dispatch application, addressed in https://github.com/vmware/dispatch/pull/143

the remaining work is moved to another issue https://github.com/vmware/dispatch/issues/158

PR #143, opened 2018-01-26
initial code - application-manager

seanhuxy:
fixes https://github.com/vmware/dispatch/issues/90


charts/dispatch/charts/application-manager/templates/ingress.yaml
```diff
@@ -0,0 +1,34 @@
...
+        paths:
+          - path: {{ .Values.ingress.path }}
+            backend:
+              serviceName: {{ include "fullname" . }}
+              servicePort: {{ .Values.service.externalPort }}
+  {{- if $tls.secretName }}
+  tls:
+    - secretName: {{ $tls.secretName }}
+      hosts:
+        - {{ default .Values.global.host .Values.ingress.host }}
```
berndtj: We/I changed the templates for how the hosts/ingress is defined.  We should be consistent.  Which could mean making all the other charts more "standard" or making this chart like the others.

seanhuxy: I will double check

charts/dispatch/charts/application-manager/templates/ingress.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if .Values.ingress.enabled -}}
+{{- $tls := default .Values.global.tls .Values.ingress.tls -}}
```
neosab: `{{- $ingress_host := default .Values.global.host .Values.ingress.host -}}`

charts/dispatch/charts/application-manager/templates/ingress.yaml
```diff
@@ -0,0 +1,34 @@
...
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    ingress.kubernetes.io/auth-signin: https://$host/oauth2/start
+    ingress.kubernetes.io/auth-url: http://{{ .Release.Name }}-oauth2-proxy.{{ .Release.Namespace }}.svc.cluster.local/oauth2/auth
+    {{- range $key, $value := .Values.ingress.annotations }}
+      {{ $key }}: {{ $value | quote }}
+    {{- end }}
+spec:
+  rules:
+    - host: {{ default .Values.global.host .Values.ingress.host }}
```
neosab: Can you add this check here:- 
      ```
{{- if $ingress_host }}
host: {{ $ingress_host }}
{{- end -}}
```

neosab:
Just a minor change and I am fine with this patch :)

PR #147, opened 2018-01-28
Remove hostname requirements and rename registry params

### LINKED ISSUE #80: Clarify documentation on docker credentials and FQDN

neosab:
The patch removes the need to specify a hostname during dispatch installation and hence adding the hostname to your local /etc/hosts. This is done by removing name based virtual routing in ingress. But the user has to specify at least the IP address of the dispatch host i.e the node address if using NodePort or the LB IP if using loadbalancer (or use hostname as before). 

The IP address is needed for generating the certs and also for setting the redirect_uri for oauth2 proxy.

The patch also renames image registry's host parameter to 'name'. This is because the term 'host' is misleading for a registry which may also contain the username or repo details as part of it. e.g gcr.io/<PROJECT_ID> or docker.io/\<username\>

Fixes #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,14 @@ The diagram below illustrates the different components which make up the Dispatc
 
 ## Quickstart
 
-Assumptions:
+### Assumptions:
 
 * Minikube is installed.  If not, simply download and install minikube
   (See [minikube](#minikube)).
 * Kubectl installed and configured for minikube.
 * Running on MacOS
 
+### Start Minikube and initialize Helm
 ```
 # Start minikube (if not already started)
 $ minikube start --vm-driver=hyperkit --bootstrapper=kubeadm --disk-size=50g --memory=6144 --kubernetes-version=v1.8.1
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -171,21 +185,23 @@ api-gateway-kongproxy   10.107.109.1   <nodes>       80:31788/TCP,443:32611/TCP
 We are looking at the port associated with https/443 => 32611
 
 ```
-$ curl -k "https://api.dispatch.local:32611/hello" -H "Content-Type: application/json" -d '{"name": "Jon", "place": "winterfell"}'
+$ curl -k "https://$DISPATCH_HOST:32611/hello" -H "Content-Type: application/json" -d '{"name": "Jon", "place": "winterfell"}'
```
seanhuxy: the host for api-gateway could be different from dispatch host (especially in a hosted k8s cluster), should we consider to differentiate them?

README.md
```diff
@@ -36,47 +37,59 @@ $ brew install kubernetes-helm
 $ helm init
 ```
 
-Add the minikube IP to /etc/hosts:
-```
-$ sudo sh -c 'echo "$(minikube ip) dispatch.local api.dispatch.local" >> /etc/hosts'
-```
+### Configure Authentication
+Configure an OAuth2 Client App at GitHub with instruction [How to Create OAuth Client
+App](https://vmware.github.io/dispatch/guides/create-oauth-client-app.html)
```
seanhuxy: when the host is optional now and we use IPs instead, what``Homepage URL`` and ``Authorization callback URL`` should be provided in github oauth2 app portal.

haven't test that with github, but you may have already.

neosab: Using IP address instead of hostname will work. I have specified it in the *Update the Callback URL of your OAuth2 Client App* section. Probably will add a note to the Oauth2 Guide as well.

README.md
neosab: The quickstart section so far only deals with minikube so wanted to keep it simple. We should elaborate the difference in standard installation.

Issue #146, opened 2018-01-27
Obfuscating credentials in dispatch install


markpeek:
- [X] Bug
- [ ] Feature
- [ ] Enhancement

## Detailed Description
Running ```dispatch install --debug``` will print out quite a lot of data. Some of this data is secret credentials that should be obfuscated for security purposes.

## Context
It is best practice not to print out credentials to prevent accidentally leaking those credentials. The use of ```--debug``` should be more careful about what it exposes. Perhaps there could be a non-obfuscation option in case that is needed for real debugging of the credentials.

berndtj:
This is going to be tough... as we are just printing out helm output.  We could try and ensure all credentials are stored as secrets and not passed in through chart configs.  This is probably a good idea, but may require that we modify dependent charts.

seanhuxy:
Agree that we should store credentials into secrets. (currently, most configurations/credentials are import from values.yaml directly or from values.yaml --> config maps --> (secrets) to the dispatch manager.

In addition, we should also consider avoid pass credentials to values.yaml and then to k8s secrets, but import directly from a file. it is pain to pass values among so many layers, and it is safer to keep the logic as simple as possible.

PR #148, opened 2018-01-29
Authz PoC

neosab:
Changes in this PR:-

1. `Oauth2Proxy` is now a sidecar container for Identity Manager deployment
   1.1 Hence, move login endpoint from `/oauth2/start` to `/v1/iam/oauth2/start`
2. Redirect all authentication requests to identity manager (handled by oauth2proxy sidecar)
3. Identity Manager's new `/v1/iam/auth` endpoint handles authorization
4. Authorization enforced with the help of[ casbin library](https://github.com/casbin/casbin)
5. Bonus:  SkipAuth is now handled on the server side
   
TODO (in later PR's)
 * Oauth2Proxy chart needs to be removed
 * Specify a default superuser in install `config.yaml`
 * Specify resource in policies (currently policies apply to all resources)
 * API to dynamically add policies

This patch is a PoC work for adding basic authorization to Dispatch. The work closely follows the proposed draft:- https://docs.google.com/document/d/1h7VULI4-uo-mrxOfqUkV3DIsgudVJe6MDnsXO5S5B1U/edit?usp=sharing


README.md
```diff
@@ -102,7 +102,7 @@ Dispatch runs on a non-standard https port on minikube since it uses
 NodePort for the ingress controller service. Hence, update the
 Authorization callback URL of your OAuth2 Client App (created by following
 [How to Create OAuth Client App](docs/create-oauth-client-app.md)) from
-`https://dispatch.local/oauth2` to `https://dispatch.local:<port>/oauth2`
+`https://dispatch.local/v1/iam/oauth2` to `https://dispatch.local:<port>/oauth2`
```
seanhuxy: do you mean `https://dispatch.local:<port>/v1/iam/oauth2` at the second quote?

charts/dispatch/charts/identity-manager/values.yaml
```diff
@@ -13,7 +13,20 @@ service:
   name: identity-manager
   type: ClusterIP
   externalPort: 80
-  internalPort: 8080
+  internalPort: 80
```
seanhuxy: saw you removed the reference to service.internalPort in some place and replace it with 8080. it is confusing you also use 80 here, is this your intention?

README.md
neosab: Oh yeah, I missed to change it. thanks :)

charts/dispatch/charts/identity-manager/values.yaml
neosab: Previously, IDM  listened on port 8080 and hence internalPort for the corresponding k8s service was also 8080.

I have now added Oauth2Proxy as a side-car which will proxy requests to Identity Manager. Hence, I need to switch internalPort to the port of Oauth2proxy (i.e standard http port). Thus, IDM's port can be directly set to 8080 since it is upstream for Oauth2proxy and not exposed to the service. Is that clear?

charts/dispatch/charts/identity-manager/templates/secret.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
berndtj: See #146. We may want to make secret definition an explicit prerequisite to chart installation.  That way no sensitive information is required to go into a values.yaml file, and we don't need to worry about credentials leaking to screen.  We can also handle this in our install command instead of relying on charts... thoughts?

pkg/identity-manager/handlers.go
```diff
@@ -72,3 +111,26 @@ func (h *Handlers) redirect(params operations.RedirectParams, principal interfac
 	location := fmt.Sprintf("%s?%s", redirect, values.Encode())
 	return operations.NewRedirectFound().WithLocation(location)
 }
+
+func getRequestAttributes(request *http.Request) *attributesRecord {
+	log.Printf("Headers: %s\n", request.Header)
```
berndtj: Debug

charts/dispatch/charts/identity-manager/values.yaml
seanhuxy: sounds great. suggest that it will be even better/clear if there's some notes/naming to clarify on 
80 --> oauth2-proxy
8080 --> IDM
as mostly people think internalPort is for the main service (i.e. IDM here)

neosab: Sure @seanhuxy Rev'ing

pkg/identity-manager/handlers.go
neosab: Done

charts/dispatch/charts/identity-manager/templates/deployment.yaml
```diff
@@ -27,9 +27,10 @@ spec:
           imagePullPolicy: {{ default .Values.global.pullPolicy .Values.image.pullPolicy }}
           args:
             - "--cookie-name=_oauth2_proxy"
+            - "--policy-file=/etc/config/policy.csv"
             - "--host=0.0.0.0"
-            - "--port={{ .Values.service.internalPort }}"
-            - "--tls-port=443"
+            - "--port=8080"
+            - "--tls-port=8443"
```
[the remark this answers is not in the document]
neosab: Thanks!

pkg/identity-manager/handlers.go
```diff
@@ -72,3 +111,26 @@ func (h *Handlers) redirect(params operations.RedirectParams, principal interfac
...
 }
+
+func getRequestAttributes(request *http.Request) *attributesRecord {
+	log.Debug("Headers: %s\n", request.Header)
+	// Retrieve the original request method and map REST verb to policy operations
+	requestMethod := request.Header.Get("X-Original-Method")
+	operation := ""
+	switch requestMethod {
+	case "GET":
+		operation = "get"
```
tenczar: If this were Java I would recommend making `operation` into an enum value. Not sure how golang handles this, but an Enum makes sure that someone doesn't misspell the operation name causing a bug in the future. Not worth bending over backwards if golang doesn't support an analogous construct.

kars7e: This could be addressed with constants from `http` package: https://golang.org/pkg/net/http/#pkg-constants

neosab: @tenczar Yeah makes sense. Thanks @kars7e I will use the http pkg constants.

neosab:
E2E Passed - Good to go!

PR #150, opened 2018-01-29
Bump dispatch chart version

neosab:
After the changes to the chart in #147 the version needs a bump.


neosab:
Not needed. It's already updated

neosab:
My bad I thought this was updated. Reopening..

berndtj:
So, I think we do want to bump the chart version after all, just in case folks don't use the install command.  Also, we need to update the image number in the charts/dispatch/values.yaml file.

berndtj:
I'm closing this as we should just bump the version next time we make changes to the charts.  We kind of missed the boat on this one

PR #156, opened 2018-01-30
[cli/api] Add update functionality for API to the cli

tenczar:
Adds update functionality for APIs into the cli.

### DIFF
--- a/pkg/api-manager/handlers.go
+++ b/pkg/api-manager/handlers.go
@@ -281,8 +281,9 @@ func (h *Handlers) updateAPI(params endpoint.UpdateAPIParams, principal interfac
 				Message: swag.String("api not found"),
 			})
 	}
-	e.Status = entitystore.StatusUPDATING
+
 	updatedEntity := apiModelOntoEntity(params.Body)
+	updatedEntity.Status = entitystore.StatusUPDATING
 	updatedEntity.API.ID = e.API.ID
 	updatedEntity.API.CreatedAt = e.API.CreatedAt
 	if _, err := h.Store.Update(e.Revision, updatedEntity); err != nil {
--- a/pkg/dispatchcli/cmd/update_api.go
+++ b/pkg/dispatchcli/cmd/update_api.go
@@ -6,12 +6,10 @@
 package cmd
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
+	"strconv"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/vmware/dispatch/pkg/api-manager/gen/client/endpoint"
 	"github.com/vmware/dispatch/pkg/api-manager/gen/models"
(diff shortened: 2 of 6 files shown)


pkg/dispatchcli/cmd/update_api.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return cmd
+}
+
+func updateAPI(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []string) error {
+	filePath := args[0]
+	api := models.API{}
+
+	if filePath != "" {
```
seanhuxy: will we only be able to update an api from file? 
I expect we are able to do quick update with some simple flags, just like ``create_api.go``

tenczar: The problem I found is that the cobra library that we are using always sets a default value for all of the flags. In that case I can't tell whether the value was set by the user or the excluded it and the value was populated by the library. I thought about maybe using a string array as the input type for each. That way I can say that the array being empty indicated that the flag was not used. In that case I'm not sure how the current update implementations would handle a mostly empty update object. Would it empty the corresponding fields in the backing store? Should the cli do a get of the actual object before doing an update...

tl;dr: If we want people to make updates to a resource via flags there are several changes that need to be made including potentially a round trip to get the update item before performing an update. If that is the desired functionality, I will add that.

seanhuxy: you are able to set default values 
``cmd.Flags().StringVarP(&functionName, "func", "f", "", "get all apis for specified function")`` the fourth one is default value "" (empty)

you should only pass flags that is set (which means not the default value) to backend.

to my best knowledge, the backend handler do a get first and then will only updated the field passed from the front-end. this need double check, but definitely you don't need to do a cli get in front-end.

berndtj: @tenczar I think I mislead you yesterday :).  I agree with @seanhuxy, though I think we may actually want to do the get and merge in the client.  This is sort of the difference between PUT/PATCH in APIs. I am however open to discussion here.

berndtj:
The comments from @seanhuxy still need to be addressed.  Additionally I'd like to see some tests in e2e that exercise this new functionality.

pkg/api-manager/handlers.go
```diff
@@ -281,8 +281,9 @@ func (h *Handlers) updateAPI(params endpoint.UpdateAPIParams, principal interfac
 				Message: swag.String("api not found"),
 			})
 	}
-	e.Status = entitystore.StatusUPDATING
+
 	updatedEntity := apiModelOntoEntity(params.Body)
+	updatedEntity.Status = entitystore.StatusUPDATING
```
berndtj: I wonder if we should use the merge struct lib for updates to make sure we are not forgetting fields to copy over: https://github.com/imdario/mergo

tenczar: I'll play with it for future work and see if it is worth porting back to this and base image.

pkg/dispatchcli/cmd/update_api.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/vmware/dispatch/pkg/api-manager/gen/client/endpoint"
+	"github.com/vmware/dispatch/pkg/api-manager/gen/models"
+	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
+)
+
+var (
+	updateAPILong    = "update an api based on json"
+	updateAPIExample = `dispatch update api my_api --path /new/path`
+
+	httpsOnlyStr string
```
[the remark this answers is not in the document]
tenczar: The ones defined in create_api.go are booleans.  A user can't do something like `dispatch update api myapi --https-only false` Cobra apparently doesn't recognize values for these boolean flags. If the flag is used then the boolean is set to true.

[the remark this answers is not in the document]
berndtj: OK... before going further on more endpoints, let's talk with the team about how this should work.  I don't think we were really thinking everything through initially.  The problem I want to avoid is now the CLI is a bit inconsistent in that on create they are flags and on update they take values.  There are other ways of addressing this, and we should explore.  For instance, kubectl has `edit` which just brings you directly into an editor of the resource's document.

PR #167, opened 2018-02-02
WIP: images are now built not just pass-through

berndtj:
This is the first step towards being able to build images with
dependencies.  We now build a docker image based on the base-image.
There are other cool things in this commit, such as being able
to use a local repository (yay!) `helm install stable/docker-registry`.

* Added docker in docker side-cars to image manager and function
  manager so that they can reach local repos within the k8s cluster
  - Set DOCKER_HOST to the localhost socket as opposed to the
    host's unix socket
* Build docker images based on the base-image
* Unify the repostory config (do we need separate repo configs for
  FaaSs and image manager?)
* Function manager is more async now as images are not necessarily
  "ready" when function created.
  - Should make "sync" smarter...
* Tried to generalize the image building, but I haven't changed the
  function manager code to use it... yet.
* Image manager now uses global config.

Testing Done:
- Installed and ran chart
- Needs more tests and e2e

### DIFF
--- a/charts/dispatch/values.yaml
+++ b/charts/dispatch/values.yaml
@@ -5,7 +5,7 @@ global:
   pullPolicy: IfNotPresent
   organization: dispatch
   image:
-    tag: v0.1.2
+    tag: v0.1.3
     host: vmware
   debug: true
   trace: false
@@ -25,5 +25,8 @@ global:
     database: dispatch
     namespace: dispatch
     release: postgres
+  registry:
+    insecure: true
+    uri: docker-docker-registry.docker.svc.cluster.local:5000
 rabbitmq:
   rabbitmqPassword: serverless
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Make a note of the `<port>` as this will be used in the next step.
 Dispatch runs on a non-standard https port on minikube since it uses
 NodePort for the ingress controller service. Hence, update the
 Authorization callback URL of your OAuth2 Client App (created by following
-[How to Create OAuth Client App](docs/create-oauth-client-app.md)) from
+[How to Create OAuth Client App](docs/_guides/create-oauth-client-app.md)) from
 `https://<DISPATCH_HOST>/oauth2` to `https://<DISPATCH_HOST>:<port>/oauth2`
 where `<port>` can be found in your dispatch config file available at
 $HOME/.dispatch/config.json.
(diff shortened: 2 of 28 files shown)


charts/dispatch/values.yaml
```diff
@@ -25,5 +25,8 @@ global:
     database: dispatch
     namespace: dispatch
     release: postgres
+  registry:
+    insecure: true
```
neosab: I prefer the default to be *insecure: false* otherwise even valid registries will be marked as insecure in dind configuration. We can expose this as a config option in DispatchConfig.ImageRegistry and let the user control it. I do understand this will be an additional config for someone doing *helm install stable/docker-registry*.

berndtj:
Passes e2e locally:
```
$ INSTALL_DISPATCH=0 ./e2e/scripts/run-e2e.sh e2e/tests
=> running clean first
 ✓ Clean all

1 test, 0 failures
=> e2e/tests/apis.bats
 ✓ Create Images for test
 ✓ Create Functions for test
 ✓ Test APIs with HTTP(S)
 ✓ Test APIs with HTTPS ONLY
 ✓ Test APIs with Kong Plugins
 ✓ Test APIs with CORS
 ✓ Cleanup

7 tests, 0 failures

=> e2e/tests/functions.bats
 ✓ Batch load images
 ✓ Create node function no schema
 ✓ Execute node function no schema
 ✓ Create python function no schema
 ✓ Execute python function no schema
 ✓ Create node function with schema
 ✓ Execute node function with schema
 - Validate schema errors (skipped: schema validation only returns null)
 ✓ Delete functions
 ✓ Cleanup

10 tests, 0 failures, 1 skipped

=> e2e/tests/events.bats
 ✓ Batch load images
 ✓ Create node function with schema
 ✓ Create subscription
 ✓ Emit event
 ✓ Delete subscription
 ✓ Cleanup

6 tests, 0 failures

=> e2e/tests/secrets.bats
 ✓ Batch load images
 ✓ Create secret
 ✓ Create function with a default secret
 ✓ Create function without a default secret
 ✓ Execute function with a default secret
 ✓ Execute function without a default secret
 - Validate invalid secret errors (skipped: secret validation only returns null)
 ✓ Update secret
 ✓ Delete secrets
 ✓ Cleanup

10 tests, 0 failures, 1 skipped

=> e2e/tests/images.bats
 ✓ Base image creation
 ✓ Image creation
 ✓ Delete images
 ✓ Delete base images
 ✓ Batch load images
 ✓ Batch delete images
 ✓ Cleanup

7 tests, 0 failures

finished
```

Issue #174, opened 2018-02-02
api-gateway - cors and OPTIONS


seanhuxy:
current API-Gateway supports cors with kong cors plugins. However, it doesn't work well with A CORS preflight request (refer to: https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request), which basically a http OPTIONS request to check if CORS protocol is understood by the server. 

- [X] Bug
- [ ] Feature
- [ ] Enhancement

## Detailed Description

when cors enable: we should not only install kong cors plugin, but also enable http OPTIONS method,
we should also enable the the transform of http query parameter to http body for OPTIONS in dispatch kong plugins, (just like what we did for HTTP GET

UPDATE: due to kong errors
```
"gateway error when adding api: kong.updatePlugin: 400, {\"config.methods\":\"\\\"OPTIONS\\\" is not allowed. Allowed values are: \\\"HEAD\\\", \\\"GET\\\", \\\"POST\\\", \\\"PUT\\\", \\\"PATCH\\\", \\\"DELETE\\\"\"}\n"
```
- need to add OPTIONS for a cors enabled apis
- but need to remove OPTIONS method when creating/updating the api cors plugin

## Complexity

- [X] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [ ] Low - Annoyance, but does not impact business or functinality
- [X] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

## Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

seanhuxy:
fixed by https://github.com/vmware/dispatch/pull/177
close

PR #177, opened 2018-02-02
fix an api-gateway bug 

seanhuxy:
fix an api-gateway bug related to cors #174 

- enable query params to http body tranform for OPTIONS method, previous only for GET
- automatically add OPTIONS method for cors-enabled api
- will not add OPTIONS method while register the corresponding kong cors plugin, this is a hard requirement from kong, as listed in the issue too.

Also: for convenience of local dev: print Kong https and http port in dispatch install output
which looks like this:
```
$ dispatch install --file config.yaml --charts-dir charts
...
Installing charts/kong helm chart
Installing charts/openfaas helm chart
Installing charts/dispatch helm chart
dispatch api-gateway is running at http port 32519 and https port 31841
Config file written to: /Users/xueyangh/.dispatch/config.json
...
```
note: the line for api-gateway port is only printed only when kong is deployed at nodePort mode

### DIFF
--- a/e2e/tests/apis.bats
+++ b/e2e/tests/apis.bats
@@ -16,8 +16,8 @@ HTTP_PORT=$(cat ${DISPATCH_CONFIG_ROOT}/config.json | jq  '.["api-http-port"]')
 : ${API_GATEWAY_HTTPS_HOST:="https://api.dev.dispatch.vmware.com:${HTTPS_PORT}"}
 : ${API_GATEWAY_HTTP_HOST:="http://api.dev.dispatch.vmware.com:${HTTP_PORT}"}
 
-# echo ${API_GATEWAY_HTTPS_HOST}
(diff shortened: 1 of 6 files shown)


e2e/tests/apis.bats
```diff
@@ -5,8 +5,19 @@ set -o pipefail
 load helpers
 load variables
 
-: ${API_GATEWAY_HTTPS_HOST:="https://api.dev.dispatch.vmware.com:${API_GATEWAY_HTTPS_PORT}"}
-: ${API_GATEWAY_HTTP_HOST:="http://api.dev.dispatch.vmware.com:${API_GATEWAY_HTTP_PORT}"}
+DISPATCH_CONFIG_ROOT=~/.dispatch
+if [ "${CI}" = true ] ; then
```
berndtj: I don't think this works in concourse, if I had to guess

PR #193, opened 2018-02-06
Update README and vCenter example

neosab:
1. The insecure flag is false by default. Need to add a note asking
users to turn it on for test environments. Also, updated the
config.json with more accurate info.

2. We need to subscribe to VmBeingCreated Event in order to trigger
the function when a VM is created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,11 +96,15 @@ cat $HOME/.dispatch/config.json
     "port": <port>,
     "organization": "",
     "cookie": "",
-    "insecure": true,
+    "insecure": false,
     "skipauth": true,
-    "Json": false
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -96,12 +96,16 @@ cat $HOME/.dispatch/config.json
     "port": <port>,
     "organization": "",
     "cookie": "",
-    "insecure": true,
```
markpeek: I agree with the below note. But don't we want insecure to be true for the laptop quick start where we are also skipping auth?

berndtj: I just posted a PR without seeing this.

PR #194, opened 2018-02-06
Add an insecure config option

berndtj:
* This allows the setting to be set from config
* If self-signed certs are generated, set insecure to true

I think this is good for the dev use-case, but we want to start integrating
with lets encrypt or similar to make real certs the default on non-dev
environments.

### DIFF
--- a/pkg/dispatchcli/cmd/install.go
+++ b/pkg/dispatchcli/cmd/install.go
@@ -109,6 +109,7 @@ type dispatchInstallConfig struct {
 	OAuth2Proxy   *oauth2ProxyConfig   `json:"oauth2Proxy,omitempty" validate:"required"`
 	TLS           *tlsConfig           `json:"tls,omitempty" validate:"required"`
 	SkipAuth      bool                 `json:"skipAuth,omitempty" validate:"omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/dispatchcli/cmd/install.go
```diff
@@ -445,11 +450,14 @@ func runInstall(out, errOut io.Writer, cmd *cobra.Command, args []string) error
...
 
 	if installService("certs") || !installDryRun {
-		err = installCert(out, errOut, configDir, config.DispatchConfig.Chart.Namespace, config.DispatchConfig.Host, config.DispatchConfig.TLS)
+		insecure, err := installCert(out, errOut, configDir, config.DispatchConfig.Chart.Namespace, config.DispatchConfig.Host, config.DispatchConfig.TLS)
 		if err != nil {
 			return errors.Wrapf(err, "Error installing cert for %s", config.DispatchConfig.Host)
 		}
-		err = installCert(out, errOut, configDir, config.APIGateway.Chart.Namespace, config.APIGateway.Host, config.APIGateway.TLS)
+		if insecure {
+			config.DispatchConfig.Insecure = insecure
```
imikushin: Can this be done unconditionally?

PR #210, opened 2018-02-08
parallelize control loop and make

berndtj:
* control loop "work" now in goroutine
* images can now be made in parallel with the -j option to make
  - make -j8 ci-images
* --timeout now an option to install to avoid/mitigate helm install timeouts

### DIFF
--- a/pkg/dispatchcli/cmd/install.go
+++ b/pkg/dispatchcli/cmd/install.go
@@ -133,6 +133,7 @@ var (
 	installDryRun     = false
 	installDebug      = false
 	configDest        = i18n.T(``)
+	helmTimeout       = 300
 )
 
 // NewCmdInstall creates a command object for the generic "get" action, which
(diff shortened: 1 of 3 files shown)


pkg/dispatchcli/cmd/install.go
```diff
@@ -160,6 +161,7 @@ func NewCmdInstall(out io.Writer, errOut io.Writer) *cobra.Command {
 	cmd.Flags().StringVar(&installChartsRepo, "charts-repo", "dispatch", "Helm Chart Repo used")
 	cmd.Flags().StringVar(&installChartsDir, "charts-dir", "dispatch", "File path to local charts (for chart development)")
 	cmd.Flags().StringVarP(&configDest, "destination", "d", "~/.dispatch", "Destination of the CLI configuration")
+	cmd.Flags().IntVarP(&helmTimeout, "timeout", "t", 300, "Timeout passed to Helm when installing charts")
```
neosab: Nit: Timeout (in seconds) passed to Helm when installing charts.

PR #212, opened 2018-02-09
be able to create/list/get entities by applications

seanhuxy:
support cli commands:
- create entity with application tags
    - dispatch create [entity] --application foo ...
- list entities by application name
    - dispatch get [entity] --application foo ...
    - dispatch get [entity] -a bar
- get/update/delete with --application flag 
    - e.g. dispatch get/update/delete [entity] [entity-name] --application <application-name>
- some refactor on secret store, add an Options arg on secretService interface
- refactor on entity store
   - filter & add an Option param for Get and List
- add e2e test for applications

unit test passed


e2e/tests/apis.bats
```diff
@@ -48,7 +48,7 @@ load variables
 }
 
 @test "Test APIs with HTTPS ONLY" {
-    run dispatch create api api-test-https-only func-nodejs6 -m POST --https-only -p /https-only -a public
+    run dispatch create api api-test-https-only func-nodejs6 -m POST --https-only -p /https-only ---authuth public
```
imikushin: `---authuth` looks like it's misspelled

e2e/tests/apis.bats
```diff
@@ -99,16 +99,16 @@ load variables
...
 
 @test "Test APIs with CORS" {
-    run dispatch create api api-test-cors func-nodejs6 -m POST -m PUT -p /cors -a public --cors
+    run dispatch create api api-test-cors func-nodejs6 -m POST -m PUT -p /cors --auth public --cors
     echo_to_log
     assert_success
     run_with_retry "dispatch get api api-test-cors --json | jq -r .status" "READY" 6 5
 
-    # contains "Access-Control-Allow-Origin: *"
+    # contains "Access-Control--authllow-Origin: *"
```
imikushin: same

e2e/tests/apis.bats
imikushin: looks like an effect of replacing `-a` with `--auth`

e2e/tests/apis.bats
```diff
@@ -99,16 +99,16 @@ load variables
...
     run_with_retry "curl -s -X PUT ${API_GATEWAY_HTTPS_HOST}/cors -k -v -d '{
             \"name\": \"VMware\",
             \"place\": \"Palo Alto\"
-        }' 2>&1 | grep -c \"Access-Control-Allow-Origin: *\"" 1 6 5
+        }' 2>&1 | grep -c \"Access-Control--authllow-Origin: *\"" 1 6 5
```
imikushin: same

e2e/tests/apis.bats
seanhuxy: thanks Ivan, will fix this a replace mistake

pkg/dispatchcli/cmd/update_secret.go
```diff
@@ -35,9 +36,12 @@ func CallUpdateSecret(input interface{}) error {
 	client := secretStoreClient()
 	secretBody := input.(*models.Secret)
 
-	params := secret.NewUpdateSecretParams()
-	params.Secret = secretBody
-	params.SecretName = *secretBody.Name
+	params := &secret.UpdateSecretParams{
```
tenczar: I'm pretty sure this will break. If you don't use the NewUpdateSecretParams() function then you need to set the default timeout yourself otherwise the update will fail with a very meaningless error. At least I noticed that when adding the update for base images.

To set the timeout you just need to add `timeout: cr.DefaultTimeout,` where `cr` is `cr "github.com/go-openapi/runtime/client"`

pkg/dispatchcli/cmd/create_api.go
```diff
@@ -51,13 +52,15 @@ func NewCmdCreateAPI(out io.Writer, errOut io.Writer) *cobra.Command {
...
+
+	cmd.Flags().StringVarP(&cmdFlagApplication, "application", "a", "", "associate with an application")
 	cmd.Flags().StringArrayVarP(&hosts, "domain", "d", []string{}, "domain names that point to your API (multi-values), default: empty")
 	cmd.Flags().StringArrayVarP(&paths, "path", "p", []string{"/"}, "paths that point to your API (multi-values), default: /")
 	cmd.Flags().StringArrayVarP(&methods, "method", "m", []string{"GET"}, "methods that point to your API, default: GET")
 	cmd.Flags().BoolVar(&httpsOnly, "https-only", false, "only support https connections, default: false")
 	cmd.Flags().BoolVar(&disable, "disable", false, "disable the api, default: false")
 	cmd.Flags().BoolVar(&cors, "cors", false, "enable CORS, default: false")
-	cmd.Flags().StringVarP(&auth, "auth", "a", "public", "specify end-user authentication method, (e.g. public, basic, oauth2), default: public")
+	cmd.Flags().StringVarP(&auth, "auth", "", "public", "specify end-user authentication method, (e.g. public, basic, oauth2), default: public")
```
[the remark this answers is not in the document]
seanhuxy: Yeah, of course

pkg/dispatchcli/cmd/update_secret.go
seanhuxy: Thanks for the info! I will do that

pkg/entity-store/postgres.go
```diff
@@ -278,25 +278,57 @@ func (p *postgresEntityStore) Update(lastRevision uint64, entity Entity) (revisi
...
 	if err != nil {
 		return errors.Wrap(err, "error getting entity from db")
 	}
-
+	if rows.Next() != false {
+		return errors.New("error getting: get more than one entity")
+	}
 	return dbToEntity(row, entity)
+
+	// sql := `SELECT * FROM entity WHERE key = $1`
```
berndtj: Should this be deleted?

seanhuxy: removed

PR #218, opened 2018-02-12
Fix golint issues

markpeek:


berndtj:
@tenczar We'll make sure that we have a page on the GitHub wiki about running lint consistently for everyone.

markpeek:
@tenczar the current check is via [gometalinter](https://github.com/alecthomas/gometalinter) which runs several different checks including [golint](https://github.com/golang/lint). However, it is running it using the --errors flag which masks warnings which a stock golint will complain about. After these are fixed I likely will introduce a separate golint pass until we decide which of the other gometalinter checks we really want.

PR #223, opened 2018-02-13
Use upstream nginx-ingress helm chart

neosab:
* Switches to upstream `stable/nginx-ingress` chart
* Delete the local fork of nginx-ingress chart
* Annotation changes to support auth in the new version of chart
* Templatize the ingress resource across subcharts
* Fix a bug in `redirect_uri` when using host_ip


kars7e:
LGTM, I've retriggered the CI to see if the API issue is intermittent.

PR #236, opened 2018-02-21
Fix helm command in minikube guide

berndtj:
The helm init command currently refers to a service account
which doesn't exist and causes tiller to fail.  This is a silent
failure which prevents dispatch from installing

### DIFF
--- a/docs/_guides/minikube.md
+++ b/docs/_guides/minikube.md
@@ -77,8 +77,7 @@ kube-system   storage-provisioner           1/1       Running   0          42s
 
 ```
 $ brew install kubernetes-helm
-# If not using the tiller service account, simply run `helm init`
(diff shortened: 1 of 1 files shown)


docs/_guides/minikube.md
```diff
@@ -78,7 +78,7 @@ kube-system   storage-provisioner           1/1       Running   0          42s
 ```
 $ brew install kubernetes-helm
 # If not using the tiller service account, simply run `helm init`
```
markpeek: Likely you should remove this comment as well. Or provide alternate paths for both.

berndtj: removed.  As long as you follow the happy path (i.e. kubeadm) you shouldn't need to worry about service accounts.

PR #240, opened 2018-02-22
Remove the need for faas specific images/base-images

berndtj:
* Add the faas specific layers at function creation time rather
  than including them in the base-image
* Use image templates (Dockerfile) to build the function images
* Create new base-images which are only OS + runtime
* Replace INTRANSIT status with CREATING and use INITIALIZED
  as the initial base state
  - keep consistent with image manager
* Reformat the global config and use camel case for the JSON
  keys (consistency)
* Add a no-op function driver for easier testability locally
* Add a file image manager option which serves canned image
  image manager responses (also for local development)

Testing Done:
- Some new unittests
- WIP


imikushin:
> Replace INTRANSIT status with CREATING

Then CREATING should NOT trigger `h.Add(e)` in 
`func (dc *DefaultController) processItem(e entitystore.Entity)` 

`entitystore.StatusCREATING` should be removed from this line: [pkg/controller/controller.go:136](https://github.com/berndtj/dispatch/blob/81620963a35bfc1ecfabb00f57fa214e543b08f9/pkg/controller/controller.go#L136)

For consistency, I'd also consider UPDATING and DELETING as "inflight" and add TO_UPDATE, TO_DELETE statuses to trigger Update and Delete operations.

berndtj:
You're right... though this is tricky as we've discussed since an entity could get stuck in CREATING state if something goes wrong.  Ugh... gotta find a clean hack now :/

berndtj:
With regards to delete... we have a separate "delete" flag on the entity which we use (sometimes).  We aren't very consistent is the main takeaway.  We need to be consistent.

We could use fields (like delete), or the "spec" field in the entity for updates...  Not sure.  But I think we are overloading status, and I don't like that.  I also don't like nondescript terms like "inflight" as you lose context as to that the final state should be.

berndtj:
Looks like the api update test is failing for me... I'll check it out.  2 in a row is quite suspect.

tenczar:
Hmmmm... for some reason updating the API takes a long time. I had to add some really long delays to the run_with_retry to get the test to run. I can look into why this update is taking so long.

berndtj:
The resync period on for the control loop in the api manager is 60 seconds.  I'm fixing that.

berndtj:
Passed e2e!

berndtj:
Let's not use reviewable... I can't just respond to your comments directly in GitHub.

berndtj:
I'll try and answer your questions via comment:

1. package.json vs. package.json.lock this is probably a better question for someone more familiar with javascript development, but it's always been my understanding that package.json is the manifest file.

2. yeah, I was thinking of getting rid of writeDir also.  I can.

3. I didn't realize that there were faas specific javascript dependencies.  Yeah, I think what you propose makes sense, I'll try and make that change.

berndtj:
ok, I think at this point, I'm not going to support package-lock.json as package.json supports pinning dependencies to version or hash.  We can always extend support later to include package-lock.json

Issue #67, opened 2017-12-18
Functions not really READY


berndtj:
Based on e2e test failures, it appears that although functions are in a READY state, they are not actually ready to run yet.

```$ dispatch get run node-hello-no-schema ccd1fac2-2b80-4732-987d-0ec5727756e7 --json
{
    "blocking": true,
    "executedTime": 1513632380,
    "finishedTime": -62135596800,
    "functionId": "35aa1618-59d2-4b78-9bff-c4ca52276b13",
    "functionName": "node-hello-no-schema",
    "input": {
        "name": "Jon",
        "place": "Winterfell"
    },
    "logs": null,
    "name": "ccd1fac2-2b80-4732-987d-0ec5727756e7",
    "reason": [
        "error running function: node-hello-no-schema: Server returned unexpected status code: 500 - Can't reach service: of-35aa1618-59d2-4b78-9bff-c4ca52276b13"
    ],
    "secrets": null,
    "status": "ERROR"
}
```

This is likely due to the way that OpenFaaS creates functions.  The k8s service associated with the function is not actually up and ready by the time the function execution is called.  We mark a function READY as soon as it's submitted to OpenFaaS.

PR #242, opened 2018-02-22
Check if the function has actually been deployed in openfaas

imikushin:
There's no function status API in OpenFaaS. When creating a function
we're making sure the function deployment is available before returning
success.

### DIFF
--- a/charts/dispatch/charts/function-manager/templates/config-map.yaml
+++ b/charts/dispatch/charts/function-manager/templates/config-map.yaml
@@ -12,11 +12,12 @@ data:
         "host": "{{ .Values.faas.openwhisk.host }}"
       },
       "openfaas": {
-        "gateway": "{{ .Values.faas.openfaas.gateway }}"
+        "gateway": "{{ .Values.faas.openfaas.gateway }}",
+        "funcNamespace": "{{ .Values.faas.openfaas.namespace }}"
       },
       "riff": {
         "gateway": "{{ .Values.faas.riff.gateway }}",
-        "func_namespace": "{{ .Values.faas.riff.namespace }}"
+        "funcNamespace": "{{ .Values.faas.riff.namespace }}"
       },
       "registry": {
         "uri": "{{ default .Values.global.registry.uri .Values.registry.uri }}",
--- a/pkg/functions/openfaas/driver.go
+++ b/pkg/functions/openfaas/driver.go
@@ -13,11 +13,18 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	docker "github.com/docker/docker/client"
 	"github.com/openfaas/faas/gateway/requests"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/vmware/dispatch/pkg/utils"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/typed/apps/v1beta2"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/vmware/dispatch/pkg/functions"
 	"github.com/vmware/dispatch/pkg/trace"
@@ -25,12 +32,17 @@ import (
 
 const (
 	jsonContentType = "application/json"
+
+	defaultCreateTimeout = 60 // seconds
 )
 
 type Config struct {
 	Gateway       string
 	ImageRegistry string
 	RegistryAuth  string
+	K8sConfig     string
+	FuncNamespace string
+	CreateTimeout *int
 }
 
 type ofDriver struct {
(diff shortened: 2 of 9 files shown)


imikushin:
Fixes #67

kars7e:
I LOVE the fact you found a way to fix this. I hope it will eventually get solved in OpenFaaS directly, maybe through some watcher to reduce the chatter. But until then, this is great!

charts/dispatch/charts/function-manager/templates/config-map.yaml
```diff
@@ -12,7 +12,8 @@ data:
         "host": "{{ .Values.faas.openwhisk.host }}"
       },
       "openfaas": {
-        "gateway": "{{ .Values.faas.openfaas.gateway }}"
+        "gateway": "{{ .Values.faas.openfaas.gateway }}",
+        "func_namespace": "{{ .Values.faas.openfaas.namespace }}"
```
berndtj: Can you use camelcase?  I see we will have a small merge conflict :).

imikushin: Sure! I'll use the names from your PR :)

pkg/functions/openfaas/driver.go
```diff
@@ -94,6 +122,25 @@ func (d *ofDriver) Create(f *functions.Function, exec *functions.Exec) error {
 		}
 		return errors.Wrapf(err, "Error performing POST request, status: %v", res.StatusCode)
 	}
+
+	// make sure the function has started
+	for i := 0; i < 10; i++ {
```
berndtj: Is 10 seconds the right timeout?

pkg/functions/openfaas/driver.go
```diff
@@ -94,6 +122,25 @@ func (d *ofDriver) Create(f *functions.Function, exec *functions.Exec) error {
...
+		}
+
+		if deployment.Status.AvailableReplicas > 0 {
+			return nil
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	return errors.Errorf("Failed to start function deployment: '%s'", getID(f.ID))
```
berndtj: I assume this will put the function in error state?

pkg/functions/openfaas/driver.go
imikushin: Fixing it right now... I'm replacing it with exponential backoff with a total max wait of about 1 minute. Would that work?

pkg/functions/openfaas/driver.go
imikushin: Correct

pkg/functions/openfaas/driver.go
berndtj: Yeah, I think that's good.  Just for kicks, let's make it configurable (max timeout).

charts/dispatch/charts/function-manager/templates/config-map.yaml
berndtj: No worries, I'm going to have a merge conflict anyway.  I can easily deal with it.

imikushin:
New and improved!

pkg/utils/backoff.go
```diff
@@ -0,0 +1,58 @@
...
+func Backoff(timeout time.Duration, f func() error) error {
+	defer trace.Trace("")()
+
+	maxTimer := time.NewTimer(timeout)
+	var err error
+
+	attempt := 0
+	sleepTime := time.Duration(rand.Int63n(int64(time.Second) / 4))
+
+	for ; ; sleepTime = sleepTime + time.Duration(rand.Int63n(int64(sleepTime+1))) {
```
kars7e: If my math is right, in worst case scenario (sleepTime randoms to 0 in line 31), this will generate ~20 attempts within first second, is that your intention?

pkg/functions/openfaas/driver.go
```diff
@@ -94,6 +132,22 @@ func (d *ofDriver) Create(f *functions.Function, exec *functions.Exec) error {
 		}
 		return errors.Wrapf(err, "Error performing POST request, status: %v", res.StatusCode)
 	}
+
+	// make sure the function has started
+	return utils.Backoff(time.Duration(d.createTimeout)*time.Second, func() error {
```
kars7e: I'm thinking whether it would be better to use kubernetes informer and just listen for updates on that deployment (with a timeout if there no event that sets available replicas), but I've never used informer with deployments so... maybe some day ;)

pkg/utils/backoff.go
imikushin: Although this scenario is very unlikely, you're right, and this is NOT my intention! I'll fix this by adding a minimum sleepTime of 250ms.

imikushin: Fixed!

pkg/functions/openfaas/driver.go
berndtj: I've used informer on custom resources... it works, but it's also a bit of a pain.  I think this is fine.  We can optimize later if we really really need to.  But openfaas should also fix this.

Issue #153, opened 2018-01-30
Add support for CloudEvents In Event manager/Event driver


kars7e:
- [ ] Bug
- [ ] Feature
- [x] Enhancement

## Detailed Description
Currently, there is no standard format for events flowing in our system. We should look at CloudEvent Spec [1] and implement it as a format of events once they get through the event driver.

[1] https://github.com/cloudevents/spec/blob/14b4e7122d026b6bac21897b8c892be2fb900ef2/spec.md

## Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [x] High - Significant architectural change or large refactor

kars7e:
Fixed in #244

Issue #66, opened 2017-12-18
*_mock.go in event-manager


seanhuxy:
there's some *_mock in event-manager, suggest to move them into a /mock folder, which is did by other managers.

kars7e:
I've removed the "good first issue" label, the reason why _mock files were in the package root directory was circular dependencies that happened when `mock` subdirectory was used. To fix it, a larger refactor is needed to ensure that the interface consumers and implementations don't live in the same package. The refactor done in #244 will fix this issue.

kars7e:
Fixed via #244

Issue #32, opened 2017-12-13
Better name handling for event subscriptions


kars7e:
Name might not be significant for subscription, thus it's not required. Currently, a combination of topic and function name is used, but this may cause conflicts and weird names. Ideas how to solve it:

* Use random names generator (a name is a combination of adjective and noun, similar to imgur, etc.)
* Use UUID as name

kars7e:
Fixed via #244

PR #244, opened 2018-02-26
Add support for CloudEvents and event driver sidecar

### LINKED ISSUE #151: Sidecar for Event Driver for MQ communication
- [ ] Bug
- [x] Feature
- [ ] Enhancement

## Detailed Description
Currently, event driver must know how to talk to the message queue to be able to function correctly. We would like to remove this limitation, and only expect event drivers to return their events to standard output. Each event driver container would be accompanied by a sidecar [1], a dedicated container deployed by dispatch in the Kubernetes pod, which would handle the Message Queue connection. Simultaneously, we could explore the possibility of two-way communication with capable event sources, by writing events to the event driver’s standard input, inspired by Concourse CI Resource concept [2]. Event Driver could then subscribe to a function (or another event), and send back an update to the original event source.

[1] Brendan Burns, The Distributed System Toolkit: Patterns for Composite Containers, http://blog.kubernetes.io/2015/06/the-distributed-system-toolkit-patterns.html
[2] Pivotal Software Inc., Implementing a resource, https://concourse.ci/implementing-resources.html

## Complexity

## Impact

- [ ] Low - Annoyance, but does not impact business or functinality
- [ ] High - Blocker

kars7e:
This change is big, take a look at [Working with events](https://github.com/kars7e/dispatch/blob/event-driver-sidecar/docs/_guides/working-with-events.md) docs and [updated spec](https://github.com/kars7e/dispatch/blob/event-driver-sidecar/docs/_specs/event-manager/event-manager.md) to help you with reviewing.
Also take a look at updated [e2e tests for events](https://github.com/kars7e/dispatch/blob/event-driver-sidecar/e2e/tests/events.bats) to see how it is used.

**Change list**:
* Added e2e tests for both event emitting, driver type registration, custom driver creation, and subscription to custom driver.
* Added `event-driver-sidecar` binary. Provisioned inside event-driver pod, serves as a gateway to Message Queue.
* Added `driverclient` - a library for easier custom driver development.
* Added `EventDriverType` - new entity in the system, used to register new driver type. Accepts name & docker image.
* Added CloudEvent support. Both `emit` API and event-driver use CloudEvent Spec for input Event.
* Added custom event drivers to examples.
* Added some docs for events.
* Subscription name can now be set by the user, will be auto-generated by CLI to human-friendly name if not specified.
* Subscription reworked. Based on CloudEvent Spec, added Event Source Type as filterable attribute in Subscription.
* Added more tests to event-manager.
* Added initial support for OpenTracing.
* Added `event` to function run payload in function manager. Added `event` field to function context, which is filled if the function is executed in reaction to an event.

**Todo** (issues to be created once this is merged):
* Revise built-in event driver types
* Implement OpenTracing in other Dispatch components
* Improve vcenter driver (add unique ID, handle all events, update govmomi)
* Add support for Kafka
* Add connection retry in RabbitMQ transport

Fixes #151 #153 #32 #66


cmd/event-driver-sidecar/main.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"os"
+
+	"github.com/vmware/dispatch/pkg/event-sidecar"
+)
+
+func main() {
+	cli := eventsidecar.NewCmd(os.Stdin, os.Stdout, os.Stderr)
+	if err := cli.Execute(); err != nil {
+		os.Exit(1)
```
berndtj: Don't you want to dump err before exiting?

pkg/dispatchcli/cmd/emit.go
```diff
@@ -46,16 +46,14 @@ func NewCmdEmit(out io.Writer, errOut io.Writer) *cobra.Command {
...
 	var payload map[string]interface{}
 	err := json.Unmarshal([]byte(emitPayload), &payload)
 	if err != nil {
 		fmt.Fprintf(errOut, "Error when parsing event payload %s\n", emitPayload)
 		return err
 	}
 	emission := &models.Emission{
-		Topic:   &eventTopic,
-		Payload: payload,
+		Event: &models.CloudEvent{},
```
berndtj: nice!

pkg/event-driver/drivers/vcenter/vcenter.go
```diff
@@ -148,19 +148,25 @@ func (d *vCenterDriver) processEvent(e types.BaseEvent) (*events.Event, error) {
...
-		ID:          uuid.NewV4().String(),
-		Body:        encoded,
-		ContentType: "application/json",
+	event := events.CloudEvent{
+		Namespace:          "vcenter.vmware.com",
+		EventType:          topic,
+		EventTypeVersion:   "0.1",
+		CloudEventsVersion: events.CloudEventsVersion,
+		SourceType:         "vcenter",
+		SourceID:           "vcenter1",
```
berndtj: is source ID unique per event type or driver instance?

pkg/event-manager/controller.go
```diff
@@ -15,8 +15,8 @@ import (
 
 // Event manager constants
 const (
-	DefaultResyncPeriod = 60 * time.Second
-	DefaultWorkerNumber = 100
+	defaultResyncPeriod = 60 * time.Second
```
berndtj: 60 is too long...

pkg/event-manager/entities.go
```diff
@@ -9,31 +9,30 @@ import (
...
-	Name string `json:"name"`
+	EventType  string   `json:"eventType"`
+	SourceType string   `json:"sourceType"`
+	SourceName string   `json:"sourceName"`
+	Function   string   `json:"function"`
+	Secrets    []string `json:"secrets,omitempty"`
 }
 
-// Driver represents a event driver, (e.g. vCenter)
+// Driver represents an event driver, (e.g. vCenter)
```
berndtj: the comment makes this sound like a driver type, not a driver instance.

pkg/event-manager/handlers_driver.go
```diff
@@ -0,0 +1,256 @@
...
+
+	"github.com/vmware/dispatch/pkg/entity-store"
+	"github.com/vmware/dispatch/pkg/event-manager/gen/models"
+	driverapi "github.com/vmware/dispatch/pkg/event-manager/gen/restapi/operations/drivers"
+	"github.com/vmware/dispatch/pkg/secret-store/gen/client/secret"
+	"github.com/vmware/dispatch/pkg/trace"
+	"github.com/vmware/dispatch/pkg/utils"
+)
+
+var builtInDrivers = map[string]map[string]bool{
```
berndtj: Do we really want to hard-code the built in drivers?  Is there a way we can load them up from a manifest?

pkg/event-manager/handlers_driver.go
```diff
@@ -0,0 +1,256 @@
...
+
+	apiKeyAuth := apiclient.APIKeyAuth("cookie", "header", "cookie")
+	secrets := make(map[string]string)
+	for _, name := range driver.Secrets {
+		resp, err := SecretStoreClient().Secret.GetSecret(&secret.GetSecretParams{
+			SecretName: name,
+			Context:    context.Background(),
+		}, apiKeyAuth)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get secret %s from secret store", name)
```
berndtj: This is where our error handling sucks.  This should be returned to the user as a 400 with a clear error message which is consistent with the swagger generated 400 responses.

pkg/event-manager/handlers_driver.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for k := range template {
+		if _, ok := driver.Config[k]; ok {
+			continue
+		}
+		if _, ok := secrets[k]; ok {
+			continue
+		}
+		return fmt.Errorf("no configuration field %s in config or secrets", k)
```
berndtj: See above.

pkg/event-manager/handlers_driver.go
```diff
@@ -0,0 +1,256 @@
...
+	return nil
+}
+
+func (h *Handlers) addDriver(params driverapi.AddDriverParams, principal interface{}) middleware.Responder {
+	defer trace.Tracef("name: %s", *params.Body.Name)()
+
+	sp, _ := utils.AddHTTPTracing(params.HTTPRequest, "EventManager.addDriver")
+	defer sp.Finish()
+
+	if err := params.Body.Validate(strfmt.Default); err != nil {
```
berndtj: I wonder how many other places we aren't actually calling "validate"...

pkg/event-manager/handlers_subscription.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if err := params.Body.Validate(strfmt.Default); err != nil {
+		return eventsapi.NewEmitEventBadRequest().WithPayload(&models.Error{
+			Code:    http.StatusBadRequest,
+			Message: swag.String(fmt.Sprintf("error validating the payload: %s", err)),
+		})
+	}
+
+	e := subscriptionModelToEntity(params.Body)
+	e.Status = entitystore.StatusCREATING
```
berndtj: I think we should standardize on StatusINITIALIZED for initial state.  There will be a bigger discussion about this, as we are overloading the status field in general (trying to cram desired and actual state into a single field).

pkg/event-sidecar/cmd.go
```diff
@@ -0,0 +1,143 @@
...
+			log.Fatalf("Error while unmarshalling configuration: %s", err)
+		}
+		if sidecarCfg.Debug {
+			log.SetLevel(log.DebugLevel)
+		} else {
+			log.SetLevel(log.InfoLevel)
+		}
+	})
+
+	cmds.PersistentFlags().String("listener-protocol", "http", "Protocol to be used by listener. One of [http, grpc, pipe] ($DISPATCH_LISTENER_PROTOCOL)")
```
berndtj: So is it one sidecar per driver?  Could we have multiple drivers writing to the same sidecar?  Could the sidecar listen on multiple protocols?

pkg/events/transport/kafka.go
```diff
@@ -0,0 +1,13 @@
+///////////////////////////////////////////////////////////////////////
+// Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+///////////////////////////////////////////////////////////////////////
+
+package transport
+
+type Kafka struct {
```
berndtj: 👍

pkg/events/cloudevent.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+const (
+	// CloudEventsVersion defines version of CloudEvent specification used in Dispatch
+	CloudEventsVersion = "0.1"
+)
+
+// CloudEvent structure implements CloudEvent spec:
+// https://github.com/cloudevents/spec/blob/b0124528486d3f6b9a247cadd68d91b44b3d3ef4/spec.md
+type CloudEvent struct {
```
berndtj: We should blog about our inclusion of cloud events.

cmd/event-driver-sidecar/main.go
kars7e: The idea here was that error is only used to indicate the return code, and the actual error message is printed by the command. But then I realized I use `log.Fatal` in command so it won't even reach this section... I will rewrite it ;)

pkg/dispatchcli/cmd/emit.go
kars7e: Right? :)

pkg/event-driver/drivers/vcenter/vcenter.go
kars7e: The SourceID here is for testing, I haven't had time yet to retrieve it from vCenter but it should be different, e.g. based on the hostname.
But this is one of the things I would like to discuss. CloudEvent spec specifies SourceID and SourceType as mandatory. They describe the source of the event.

Now, the driver is created with a certain name. It's possible that two drivers (with different names) will be created for the same source (e.g. same vcenter). 

The driver is expected to feed CloudEvent-formatted to the sidecar. it picks the SourceID and sends to Sidecar. Now, sidecar must decide how to route that event. My idea was that event is routed based on 3 things: SourceType (e.g. vcenter), SourceID (e.g. vcenter1.corp.local), and EventType (e.g. `vm.created`). To ensure uniqueness in routing, I was thinking about overwriting SourceID with the name of the driver - this name is exposed to the sidecar at the time it's provisioned, and would ensure uniqueness (as name are unique per org, and different org would use different exchange in RabbitMQ). However, this would cause original SourceID coming from vcenter to be lost.
I think the best would be if I could make the routing decision based on the driver name, but somehow convey this without overwriting Source ID, so that when a function receives the event, Source ID remains the same. I will look into that.

pkg/event-manager/controller.go
kars7e: this is only name change, value is the same ;) I can update that of course.

pkg/event-manager/entities.go
kars7e: Good point, I will update it.

pkg/event-manager/handlers_driver.go
kars7e: Well, there are built-in, so... What would be the benefit of using manifest here?

pkg/event-manager/handlers_driver.go
kars7e: Yeah. Also each service will use different approach for this... Can we figure out a unified way of handling errors? Is it even possible? Is defining global error types helpful?

pkg/event-manager/handlers_driver.go
kars7e: Yeah :/ during the security review we should look at our approach to validation & consistency as well...

pkg/event-manager/handlers_subscription.go
kars7e: I don't think each type can have exactly the same set of statuses (looking at kubernetes, each type has different status set, e.g. it makes sense for Pod to have `ImagePullBackOff` state, which makes no sense for deployment), but `StatusINITIALIZED` sounds good for the initial state.

pkg/event-sidecar/cmd.go
kars7e: You got me thinking with that one. Making sidecar to listen using all protocols at once sounds reasonable, that would solve the "mode" flag in driver type.

The idea behind sidecar is that each driver gets its own. Each driver instance is then a pod, and can communicate over local TCP/IP stack (or files on shared volume, or IPC).
Now, updating single pod with more driver containers does not sound reasonable, as it would have to recreated every time (adding new containers to running pod is not allowed per: https://github.com/kubernetes/kubernetes/issues/37838). 
If we used separate pods, we would lose the ability to use localhost or shared volume. We could only use networking stack, and we would get into service discovery of where the sidecar/service lives. Also, sidecar encapsulates privileges/message routing, so that event driver doesn't have to. Because we have one sidecar per event driver, we can configure the routing/privileges at the time of driver pod deployment and not worry about it later.

However, I'm seeing some advantages from performance perspective. Maybe we could investigate having one sidecar (well, not sidecar anymore, let's call it event router perhaps) per tenant. This would ensure the routing/privileges stay the same.

pkg/events/cloudevent.go
kars7e: Definitely!

charts/dispatch/charts/event-manager/templates/deployment.yaml
```diff
@@ -39,9 +39,10 @@ spec:
             - "--function-manager={{ .Release.Name }}-function-manager"
             - "--secret-store={{ .Release.Name }}-secret-store"
             # TODO: Read password from secret
-            - "--amqpurl=amqp://user:serverless@{{ .Release.Name }}-rabbitmq:5672/"
+            - "--rabbitmq-url=amqp://user:serverless@{{ .Release.Name }}-rabbitmq:5672/"
```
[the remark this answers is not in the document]
kars7e: not really, this is more to standardize the attribute naming (it was rabbitmq in some places and amqp in others). Note that we're using amqp library for 0.9.1 version with extensions for RabbitMQ. Although technically possible, I don't think it would be feasible to support other AMQP implementations than RabbitMQ.
 
Also, the plan is to add Kafka support, so there will be another flag for that too.

docs/_guides/custom-event-drivers.md
```diff
@@ -0,0 +1,39 @@
...
+```
+The above command registers new driver type `timer-driver`, using `kars7e/timer-driver` image from Docker Hub.
+
+Now, you can create the actual event-driver:
+
+```
+dispatch create event-driver timer-driver --name timer-driver1
+```
+
+And create a subscription:
```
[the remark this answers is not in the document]
kars7e: It is WIP but please do nitpick grammar/typos!

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,121 @@
...
+```
+dispatch get event-driver vcenter1.corp.local
+```
+
+When status changes to `RUNNING`, event driver is ready to use.
+
+Event driver by itself doesn't do much. it produces events which are ingested by Dispatch, but Dispatch does not know what
+to do with them yet. To tell Dispatch to react to events, you need to create a subscription.
+
+### Event driver event types
```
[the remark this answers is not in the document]
kars7e: And it indeed is, will fix it. thanks!

pkg/event-driver/drivers/vcenter/vcenter.go
```diff
@@ -148,19 +148,25 @@ func (d *vCenterDriver) processEvent(e types.BaseEvent) (*events.Event, error) {
...
 
-	event := events.Event{
-		Topic:       topic,
-		ID:          uuid.NewV4().String(),
-		Body:        encoded,
-		ContentType: "application/json",
+	event := events.CloudEvent{
+		Namespace:          "vcenter.vmware.com",
+		EventType:          topic,
+		EventTypeVersion:   "0.1",
```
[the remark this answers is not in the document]
kars7e: EventTypeVersion is per event source, but I agree it should be a constant in the driver.

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
+---
+layout: default
+---
+
+# Working with Events in Dispatch
+
+Dispatch is designed to accept events from different sources. To be able to do that, Dispatch introduces a concept of event driver.
+Event Driver an image that knows how to ingest events from particular source. For example, if you would like to run your functions
```
tenczar: Just take the edits from here https://github.com/vmware/dispatch/pull/244/files#diff-3c7f82c0529b4f417e37fcaa6b9c0706R8 and replace the first line :)

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+in reaction to events coming from VMware vCenter, you would create an event driver of type `vcenter`. There are built-in types,
+you can also register your own type by providing a specially crafted image.
+
+## Implementation
+
+Event drivers are implemented as Kubernetes Pods. When new event driver is created, a new Kubernetes pod is deployed,
```
tenczar: When _a_ new event driver

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+with configuration matching the one provided in the request. Pod is wrapped in the deployment, so that the pod will be
```
tenczar: _The_ Pod is wrapped in _a_ deployment

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+rescheduled/restarted if needed. See [Registering custom event driver](custom-event-drivers.md) for more details.
+
+## How to use event driver in your environment
+
+### Adding event driver
+
+You can use Dispatch CLI or API to add an event driver (UI support coming soon). In following example, we're adding
```
tenczar: In _the_ following example, we're adding _a_ new vCenter event driver

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+### Adding event driver
+
+You can use Dispatch CLI or API to add an event driver (UI support coming soon). In following example, we're adding 
+new vCenter event driver. 
+
+```
+dispatch create event-driver vcenter --name vcenter1.corp.local
+```
+
+The above will create new event driver of type `vcenter` and --name `vcenter1.corp.local`. Creating an event-driver
```
tenczar: create _a_ new event driver

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+takes a while - monitor its status by running:
+
+```
+dispatch get event-driver vcenter1.corp.local
+```
+
+When status changes to `RUNNING`, event driver is ready to use.
```
tenczar: When _the_ status changes to \`RUNNING\`, _the_ event driver is ready to use.

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+
+Event driver by itself doesn't do much. it produces events which are ingested by Dispatch, but Dispatch does not know what
+to do with them yet. To tell Dispatch to react to events, you need to create a subscription.
+
+## Working with subscriptions
+### Adding subscription
+
+Subscription tells Dispatch to match certain events to certain function, so that when events matching defined conditions arrive
```
tenczar: certain _functions_

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+to the system, these functions will be executed. Events that do not match any subscriptions are automatically dropped.
+
+To add a new subscription for `vm.being.created` event in vcenter event driver, run:
```
tenczar: for _the_ \`vm.being.created\` event  / in _a_ vcenter event driver

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+The above command will print output similar to the following:
+
+```
+           NAME          | SOURCE TYPE |    EVENT TYPE    | FUNCTION NAME | STATUS |         CREATED DATE
+---------------------------------------------------------------------------------------------------------
+  complete-cicada-410962 | *           | vm.being.created | myFunctio     | READY  | Fri Dec 31 17:18:54 PST -0001
+```
+
+The above subscription will cause dispatch to execute function `myFunction` for every event of type `vm.being.created`. 
+Note the asterisk in  `SOURCE TYPE` column. This is default value and say "match all source Types".
```
tenczar: This is _the_ default value and _means_ \"match all source Types".

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Event driver event types
+
+To find out the list of event types produced by built-in event drivers, see [Built-in Event Drivers](built-in-event-drivers.md).
+
+
+## Emitting events via CLI or API
+
+Event driver is a powerful concept, but there are cases when you want to emit an event by your self or through some external application.
+Dispatch exposes dedicated API to emit events.
```
tenczar: Dispatch exposes _a_ dedicated API to emit events.

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+
+To emit an event using CLI, you can use `dispatch emit` subcommand.
```
tenczar: you can use _the_ \`dispatch emit\` subcommand

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+
+In a following example, We emit an event of type `my.event`:
```
tenczar: In _the_ following example.

This is one article change that should be made to avoid confusion.

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+dispatch emit my.event --data '{ "example": "payload"}'
+```  
+
+Events in Dispatch follow [Cloud Events specification](https://github.com/cloudevents/spec/blob/460a90d7a69f6257246487e37746797aa2ae919f/spec.md).
+There are certain attributes in Cloud Events that are mandatory. when emitting them through CLI, CLI will pick reasonable defaults.
```
tenczar: through _the_ CLI, _the_ CLI will pick reasonable defaults.

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+* `--event-id` - Event ID, should be unique within the scope of producer. Defaults to generated UUIDv4. 
+  NOTE: If source ID is `dispatch`, event-id will ALWAYS be overwritten by Dispatch and set to random UUIDv4. 
+
+You can also provide following optional details about the emitted event:
+* `--event-type-version` - Version of the event, specific to the source. Defaults to empty string.
+* `--schema-url` - URL to schema describing event data. Defaults to empty string.
+* `--content-type` - Content Type of optional data payload. Defaults to `application/json` when using `--data`, or automatically
+detected type when using `--data-from-file`. If specified, always takes precedence over detected type.
+
+You can provide data payload for Event using `--data` parameter. `--data-from-file` will read the content from specified
```
tenczar: from _the_ specified file path. Use _the_ \`binary\` switch to =

docs/_guides/working-with-events.md
```diff
@@ -0,0 +1,109 @@
...
+file path. Use `--binary` switch to automatically encode the data using base64. If dispatch CLI detects binary type,
+the content will automatically be base64-encoded.
+
+TODO: Add API example.
```
tenczar: Are you going to do this now? Should we take this out before we publish the PR? Same thing for the TODO a few lines down.

imikushin:
This PR is pretty huge. That's my only complain right now.

docs/_specs/event-manager/event-manager.md
```diff
@@ -0,0 +1,84 @@
+---
+layout: default
+---
+# Event manager
+
+Event manager handles events traffic within Dispatch. It connects events to functions using subscriptions,
```
tenczar: manager handles _event_ traffic

docs/_specs/event-manager/event-manager.md
```diff
@@ -0,0 +1,84 @@
...
+## Problem statement
+
+Functions are commonly associated with events and are executed in response to them. To effectively manage their
+relation, we need a way to handle events lifecycle (CRUD) as well as subscriptions and triggering.
+
+### User stories (examples)
+
+1. As a **developer**, I want to trigger an **event** by providing its **name/topic** and JSON-formatted **payload**, which will
+   in turn cause execution of one or more functions **subscribed** to that event.
+2. As a **developer**, I want to create a **subscription** between **event** and **function**. When event is triggered,
```
tenczar: between _an_ \*\*event\*\* and _a_ \*\*function\*\*. When _an_ event is triggered,

docs/_specs/event-manager/event-manager.md
```diff
@@ -0,0 +1,84 @@
...
+   function(s) is/are executed with event payload.
+3. As a **developer/administrator**, I want to provide sustainable way of ingesting events
```
tenczar: provide _a_ sustainable way

docs/_specs/event-manager/event-manager.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Proposed Solution
+
+An Event Manager component, responsible for event and subscription lifecycle.
+
+Event manager:
+* Exposes API for CRUD operations on subscriptions.
+* Exposes API for CRUD operations on event drivers and heir types.
```
tenczar: event drivers and _their_ types. (I believe that is what you meant)

e2e/tests/events.bats
```diff
@@ -10,36 +10,66 @@ load variables
...
     echo_to_log
     assert_success
 
-    run_with_retry "dispatch get subscription test_event_node-hello-subscribe-${RUN_ID} --json | jq -r .status" "READY" 4 5
+    run_with_retry "dispatch get subscription testsubscription --json | jq -r .status" "READY" 4 5
 }
 
 @test "Emit event" {
-    run dispatch emit test.event --payload='{"name": "Jon", "place": "Winterfell"}'
+    sleep 5 # Still seeing the issue https://github.com/vmware/dispatch/issues/67
```
tenczar: You've been removing this everywhere. Why add it here?

examples/nodejs6/debug.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function (context, params) {
```
tenczar: This seems really useful. Do we really want to bury it in the examples?

pkg/dispatchcli/cmd/create_subscription.go
```diff
@@ -23,41 +23,45 @@ var (
...
 			CheckErr(err)
 		},
 	}
-
 	cmd.Flags().StringVarP(&cmdFlagApplication, "application", "a", "", "associate with an application")
 	cmd.Flags().StringArrayVar(&createSubscriptionSecrets, "secret", []string{}, "Function secrets, can be specified multiple times or a comma-delimited string")
+
+	cmd.Flags().StringVar(&createSubscriptionName, "name", "", "Subscription name. If not specified, will be randomly generated.")
+	cmd.Flags().StringVar(&createSubscriptionEventType, "event-type", "*", "Event Type to filter on.")
+	cmd.Flags().StringVar(&createSubscriptionSourceType, "source-type", "*", "Source type to filter on. Most often it will be your event driver type.")
```
tenczar: You should add these into the use field.

pkg/dispatchcli/cmd/emit.go
```diff
@@ -41,21 +58,45 @@ func NewCmdEmit(out io.Writer, errOut io.Writer) *cobra.Command {
...
+	cmd.Flags().StringVar(&emitEventNamespace, "namespace", emitEventNamespace, "Event namespace.")
+	cmd.Flags().StringVar(&emitEventSourceID, "source-id", emitEventSourceID, "Event source ID.")
+	cmd.Flags().StringVar(&emitEventSourceType, "source-type", emitEventSourceType, "Event source type.")
+	cmd.Flags().StringVar(&emitEventID, "event-id", "", "Event ID. If empty, will be autogenerated.")
+	cmd.Flags().StringVar(&emitEventContentType, "content-type", "", "Content type of the payload. Defaults to application/json if --data is used. For --data-from-file, it will try to detect the type.")
+	cmd.Flags().StringVar(&emitEventData, "data", "", "Event data payload. Mutually exclusive with --data-from-file")
+	cmd.Flags().StringVar(&emitEventDataFromFile, "data-from-file", "", "Read event payload from the given file path. Mutually exclusive with --data")
+	cmd.Flags().BoolVar(&emitEventDataBinary, "binary", false, "Sets payload as binary (will be base64 encoded). If not specified, automatic detection will be attempted.")
+	cmd.Flags().StringVar(&emitEventSchemaURL, "event-schema-url", "", "Event Schema URL.")
```
tenczar: You should add all of these flags to the "use" field

pkg/dispatchcli/cmd/get_eventdrivertype.go
```diff
@@ -0,0 +1,118 @@
...
+			if len(args) == 1 {
+				err = getEventDriverType(out, errOut, cmd, args)
+			} else {
+				err = getEventDriverTypes(out, errOut, cmd)
+			}
+			CheckErr(err)
+		},
+	}
+	cmd.Flags().StringVarP(&cmdFlagApplication, "application", "a", "", "filter by application")
+	cmd.Flags().BoolVar(&getEventDriverTypeShowBuiltIn, "show-builtin", false, "Include built-in driver types in result")
```
tenczar: You should add these to the "Use" field

e2e/tests/events.bats
```diff
@@ -10,36 +10,66 @@ load variables
...
 
 @test "Delete subscription" {
-    run dispatch delete subscription test_event_node-hello-subscribe-${RUN_ID}
+    run dispatch delete subscription testsubscription
     echo_to_log
     assert_success
 }
 
+@test "Register new event driver type" {
+    run dispatch create eventdrivertype ticker kars7e/timer:latest
```
berndtj: I'll throw this on vmware docker hub.  Remind me.

pkg/dispatchcli/cmd/cmd.go
```diff
@@ -34,7 +35,8 @@ var validResources = i18n.T(`Valid resource types include:
 	* apis
 	* applications
 	* base-images
-	* event-drivers
+	* eventdrivers
+    * eventdrivertypes
```
berndtj: nit, weird formatting

pkg/dispatchcli/cmd/delete_eventdrivertype.go
```diff
@@ -0,0 +1,79 @@
...
+
+// NewCmdDeleteEventDriverType creates command responsible for deleting EventDriverType.
+func NewCmdDeleteEventDriverType(out io.Writer, errOut io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "eventdrivertype TYPE_NAME",
+		Short:   i18n.T("Delete event driver type"),
+		Long:    deleteEventDriverTypeLong,
+		Example: deleteEventDriverTypeExample,
+		Args:    cobra.ExactArgs(1),
+		Aliases: []string{"eventdrivertypes", "event-driver-type", "event-driver-types"},
```
berndtj: how about eventdriver-type also :)

pkg/dispatchcli/cmd/get_subscription.go
```diff
@@ -91,11 +91,11 @@ func formatSubscriptionOutput(out io.Writer, list bool, subscriptions []*models.
 		return encoder.Encode(subscriptions[0])
 	}
 	table := tablewriter.NewWriter(out)
-	table.SetHeader([]string{"Name", "Topic", "Subscriber Type", "Subscriber Name", "Status", "Created Date"})
+	table.SetHeader([]string{"Name", "Source Type", "Event Type", "Function name", "Status", "Created Date"})
```
berndtj: why lowercase name? seems inconsistent.

kars7e:
Thank you folks for the effort, I know I didn't make it easy. I'll try to stop myself earlier with the next PR!  Since the most recent comments are typos/comment enhancements, I will merge it to get this monster out of the way and address the comments in a follow-up patch.

e2e/tests/events.bats
kars7e: This is the only place where, after removing the sleeps, I've noticed the issue again.

examples/nodejs6/debug.js
kars7e: I'm not sure where else we could place it, would like this to be moved to tests directly?

PR #246, opened 2018-03-01
Add Policy API's

neosab:
* Adds Policy API's `GET, POST, DELETE`
* Adds a simple casbin storage adapter for loading policy entities
* Adds controller and policy entity handlers
* Adds support for `--enable-bootstrap-mode` for identity-manager (to add initial policies)
* Ability to specify a bootstrap user in install config.yaml
* some other necessary ingress changes

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/246)


charts/dispatch/charts/identity-manager/templates/deployment.yaml
```diff
@@ -28,12 +28,19 @@ spec:
           imagePullPolicy: {{ default .Values.global.pullPolicy .Values.image.pullPolicy }}
           args:
             - "--cookie-name=_oauth2_proxy"
-            - "--policy-file=/etc/config/policy.csv"
             - "--host=0.0.0.0"
             - "--port=8080"
+            - "--db-file={{ default .Release.Name .Values.global.db.release }}-{{ .Values.global.db.host }}.{{ default .Release.Namespace .Values.global.db.namespace }}:{{ .Values.global.db.port }}"
```
berndtj: What do we want to make a CLI arg vs. what goes in config.json?  I want us to be consistent across services.  I just took things out of CLI args for function manager.  I don't really care one way or the other, I just want us to be consistent.

charts/dispatch/charts/identity-manager/templates/ingress.noauth.yaml
```diff
@@ -4,23 +4,29 @@
...
   annotations:
-    {{- range $key, $value := .Values.ingress.home.annotations }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
     nginx.ingress.kubernetes.io/configuration-snippet: |
       error_page 403 = @403.json;
+    nginx.ingress.kubernetes.io/server-snippet: |
+      location @403.json {
+        default_type application/json;
+        return 403 '{"code": 403,"message": "Forbidden, please make sure you are logged in and have required privileges"}';
```
berndtj: 👍

charts/dispatch/charts/identity-manager/templates/secret.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     heritage: "{{ .Release.Service }}"
 type: Opaque
 data:
+  bootstrap_user: {{ default "" .Values.bootstrapUser | trim | b64enc | quote }}
```
berndtj: nit, inconsistent spacing

pkg/identity-manager/casbin_entity_adapter.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func NewCasbinEntityAdapter(store entitystore.EntityStore) *CasbinEntityAdapter {
+	return &CasbinEntityAdapter{store: store}
+}
+
+func (a *CasbinEntityAdapter) LoadPolicy(model model.Model) error {
+	var policies []*Policy
+
+	// Policy status like CREATING, DELETING doesn't have any effect on
```
tenczar: Doesn't have any effect on what?

pkg/identity-manager/policy_handler.go
```diff
@@ -0,0 +1,87 @@
+///////////////////////////////////////////////////////////////////////
```
tenczar: This kind of confused me. The api-manager and image-manager defined this handler in the same controller.go file as the associated controller. We probably want to keep all of these consistent. I think the way you did it here makes more sense.

cmd/identity-manager/main.go
```diff
@@ -82,9 +84,28 @@ func main() {
...
+	es, err := entitystore.NewFromBackend(
+		entitystore.BackendConfig{
+			Backend:  identitymanager.IdentityManagerFlags.DbBackend,
+			Address:  identitymanager.IdentityManagerFlags.DbFile,
+			Bucket:   identitymanager.IdentityManagerFlags.DbDatabase,
+			Username: identitymanager.IdentityManagerFlags.DbUser,
+			Password: identitymanager.IdentityManagerFlags.DbPassword,
+		})
+	if err != nil {
+		log.Fatalln(err)
```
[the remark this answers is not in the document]
kars7e: `log.Fatalln` runs `os.Exit(1)`, so it should bail out here

pkg/identity-manager/casbin_entity_adapter.go
```diff
@@ -0,0 +1,72 @@
...
+	opts := entitystore.Options{
+		Filter: entitystore.FilterExists(),
+	}
+
+	log.Debug("Reloading policies")
+	err := a.store.List(IdentityManagerFlags.OrgID, opts, &policies)
+	for _, policy := range policies {
+		// Casbin authorization rules are of the form (subject, resource, action) and hence the need to iterate over all rule fields.
+		for _, rule := range policy.Rules {
```
kars7e: That's a lot of `for`s! :)

pkg/identity-manager/casbin_entity_adapter.go
```diff
@@ -0,0 +1,72 @@
...
+
+type CasbinEntityAdapter struct {
+	store entitystore.EntityStore
+}
+
+func NewCasbinEntityAdapter(store entitystore.EntityStore) *CasbinEntityAdapter {
+	return &CasbinEntityAdapter{store: store}
+}
+
+func (a *CasbinEntityAdapter) LoadPolicy(model model.Model) error {
```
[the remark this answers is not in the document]
neosab: Good point, I will alias it.

pkg/identity-manager/casbin_entity_adapter.go
neosab: I guess I context switched while typing that ;), I will fix it. Basically, I meant that LoadPolicy doesn't inspect the status of the policy entity. The status is more for the controllers to act upon.

pkg/identity-manager/casbin_entity_adapter.go
neosab: I KNEW someone would say that, anyway I will fix it with some memoization 📝

pkg/identity-manager/policy_handler.go
neosab: Hmm, I kinda followed the suite from event manager. While working on this I noticed a lot of minor inconsistencies within different microservices. We need to fix them.

charts/dispatch/charts/identity-manager/templates/deployment.yaml
neosab: I still see command args being passed to the function manager's deployment https://github.com/vmware/dispatch/blob/master/charts/dispatch/charts/function-manager/templates/deployment.yaml#L31

Did you mean something else?

charts/dispatch/charts/identity-manager/templates/secret.yaml
neosab: Will fix it thanks!

cmd/identity-manager/main.go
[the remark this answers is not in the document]
tenczar: Not sure how I feel about that. I guess it is convenient, but the doc comment should really mention that it calls os.Exit(1).

Issue #98, opened 2018-01-09
Creating a function with name that already exists returns 500 error.


kars7e:
- [X] Bug
- [ ] Feature
- [ ] Enhancement

## Detailed Description
When creating a function, if function by this name already exists, CLI returns 500 error with a meaningless error message:
```
$ dispatch create function powercli powercli-list-vms examples/powershell/powercli-list-vms.ps1
Created function: powercli-list-vms

$ dispatch create function powercli powercli-list-vms examples/powershell/powercli-list-vms.ps1
[Code: 500] Error: internal server error when storing a new function
```

It should instead return a meaningful error message.

## Complexity

- [X] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [X] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

PR #263, opened 2018-03-05
Appropriate error when creating an entity with non-unique name

imikushin:
When trying to create a function, return 409 Conflict if a function
with the same name already exists.


imikushin:
Fixes #98

PR #273, opened 2018-03-05
Add full API model support to create, update, delete by file

tenczar:


pkg/dispatchcli/cmd/create.go
```diff
@@ -67,7 +69,18 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 			return errors.Wrapf(err, "Error decoding document %s", string(doc))
 		}
 		switch docKind := k.Kind; docKind {
-		case "base-image":
+		case "API":
```
imikushin: String value `"API"` is the same as `APIKind` const above. 

Would you consider moving the `*Kind` constants to a package which can be imported everywhere they are needed, so constants could be used instead of string values?

swagger/api-manager.yaml
```diff
@@ -220,6 +220,9 @@ definitions:
       name:
         type: string
         pattern: '^[\w\d\-]+$'
+      kind:
```
kars7e: `kind` attribute should have `readOnly` attribute set to `true` (same for other swagger files)

pkg/secret-store/builder/vmw_secret_builder.go
```diff
@@ -14,6 +14,9 @@ import (
 	"k8s.io/api/core/v1"
 )
 
+// SecretKind a constant representing the kind of the Secret model
+const SecretKind = "Secret"
+
 // VmwSecretBuilder type
 type VmwSecretBuilder struct {
```
kars7e: not related to this PR, but seems like this name should be changed to dispatchSecretBuilder

pkg/dispatchcli/cmd/create.go
tenczar: I think this is a great idea. Do you have a recommendation of where I should create this constants package?

pkg/utils/constants.go
```diff
@@ -0,0 +1,25 @@
+package utils
```
imikushin: We need the license header here: CI is complaining about it ;)

Make sure gofmt is happy too.

kars7e:
e2e is failing on unrelated issue, I will try to track it down and apply some fix. this change can go through.

pkg/dispatchcli/cmd/delete.go
```diff
@@ -38,10 +38,10 @@ func NewCmdDelete(out io.Writer, errOut io.Writer) *cobra.Command {
 			}
 
 			deleteMap := map[string]modelAction{
-				"image":      CallDeleteImage,
-				"base-image": CallDeleteBaseImage,
-				"function":   CallDeleteFunction,
-				"secret":     CallDeleteSecret,
+				"Image":     CallDeleteImage,
```
neosab: utils.ImageKind and similar for keys below

tenczar: Yeah, I just noticed this in working on my CallUpdateImage changes. I have noted this to be fixed later. thanks.

neosab: Np!

PR #276, opened 2018-03-05
Return 409 Conflict on non-unique names

imikushin:
Modify APIs to return 409 Conflict on non-unique names where
appropriate.

### DIFF
--- a/pkg/event-manager/subscriptions/http_handlers.go
+++ b/pkg/event-manager/subscriptions/http_handlers.go
@@ -18,7 +18,6 @@ import (
 	"github.com/vmware/dispatch/pkg/entity-store"
 	"github.com/vmware/dispatch/pkg/event-manager/gen/models"
 	"github.com/vmware/dispatch/pkg/event-manager/gen/restapi/operations"
-	eventsapi "github.com/vmware/dispatch/pkg/event-manager/gen/restapi/operations/events"
 	subscriptionsapi "github.com/vmware/dispatch/pkg/event-manager/gen/restapi/operations/subscriptions"
 	"github.com/vmware/dispatch/pkg/event-manager/subscriptions/entities"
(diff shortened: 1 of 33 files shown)


pkg/event-manager/subscriptions/http_handlers.go
```diff
@@ -63,7 +62,7 @@ func (h *Handlers) addSubscription(params subscriptionsapi.AddSubscriptionParams
 	defer sp.Finish()
 
 	if err := params.Body.Validate(strfmt.Default); err != nil {
-		return eventsapi.NewEmitEventBadRequest().WithPayload(&models.Error{
```
imikushin: Looks like this was here by mistake

pkg/event-manager/subscriptions/http_handlers.go
```diff
@@ -74,8 +73,14 @@ func (h *Handlers) addSubscription(params subscriptionsapi.AddSubscriptionParams
...
 	_, err := h.store.Add(s)
 	if err != nil {
+		if entitystore.IsUniqueViolation(err) {
+			return subscriptionsapi.NewAddSubscriptionConflict().WithPayload(&models.Error{
+				Code:    http.StatusConflict,
+				Message: swag.String("error creating subscription: non-unique name"),
+			})
+		}
 		log.Errorf("error when storing the subscription: %+v", err)
-		return eventsapi.NewEmitEventInternalServerError().WithPayload(&models.Error{
```
imikushin: Looks like this was here by mistake

pkg/event-manager/subscriptions/http_handlers.go
kars7e: Good catch! It would be great if we could unify our models/error responses... maybe some day :)

imikushin:
This is a follow-up on #263

PR #284, opened 2018-03-07
Support single ns mode and better service selection during install

kars7e:
**Changes:**
* Add `--single-namespace` flag to `dispatch install` to support single namespace installation
* `--service` install flag reworked. Maintains backward compatibility. You can specify `no-` prefix for any service to disable its installation, e.g. `dispatch install --service no-ingress` will install everything except ingress.
* explicit namespace existence check to avoid corner cases where `create namespace` returns "Forbidden" instead of "AlreadyExists" even if a namespace already exists (due to RBAC settings).
* Added docker registry to uninstall, also added `--keep-namespaces` flag to avoid errors in environments with precreated namespaces
* Added support for domain names longer than 64 characters (previously there was an issue with certificate)
* Fixed some linting errors

### DIFF
--- a/pkg/dispatchcli/cmd/install.go
+++ b/pkg/dispatchcli/cmd/install.go
@@ -21,10 +21,10 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	validator "gopkg.in/go-playground/validator.v9"
+	"gopkg.in/go-playground/validator.v9"
 
 	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
 )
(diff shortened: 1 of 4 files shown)


pkg/dispatchcli/cmd/install.go
```diff
@@ -21,10 +21,10 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/mitchellh/go-homedir"
```
berndtj: Is this a lint issue?  Because I kind of like the explicit package name where the name differs from the package URI.

pkg/dispatchcli/cmd/install.go
```diff
@@ -172,49 +188,83 @@ func NewCmdInstall(out io.Writer, errOut io.Writer) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&installConfigFile, "file", "f", "", "Path to YAML file")
-	cmd.Flags().StringArrayVarP(&installServices, "service", "s", []string{}, "Service to install (defaults to all)")
+	cmd.Flags().StringArrayVarP(&installServices, "service", "s", []string{}, "Service to install (defaults to all). Add 'no-' prefix to service name to disable it.")
 	cmd.Flags().BoolVar(&installDryRun, "dry-run", false, "Do a dry run, but don't install anything")
 	cmd.Flags().BoolVar(&installDebug, "debug", false, "Extra debug output")
 	cmd.Flags().StringVar(&installChartsRepo, "charts-repo", "dispatch", "Helm Chart Repo used")
 	cmd.Flags().StringVar(&installChartsDir, "charts-dir", "dispatch", "File path to local charts (for chart development)")
+	cmd.Flags().StringVar(&installSingleNS, "single-namespace", "", "If specified, all dispatch components will be installed to that namespace")
```
berndtj: Do we need to prefix this option with "single"?  Also, should we install into a single namespace (dispatch) by default?

pkg/dispatchcli/cmd/install.go
kars7e: It was highlighted for me (might not be from golint but some other linter).

pkg/dispatchcli/cmd/install.go
kars7e: I added the prefix to clearly indicate that the deployment model differs from the default one (multiple namespaces). If we switch to single namespace being default, using just `--namespace` makes more sense. but that's a bigger change (we need to change CI as well), so let's do it in a separate patch.

PR #285, opened 2018-03-07
Cleanup leftover resources after deleting a cluster on GKE

rjew:
Leftover loadbalancers and persistent storage are removed after e2e CI pipeline is complete

### DIFF
--- a/ci/e2e/cleanup.yml
+++ b/ci/e2e/cleanup.yml
@@ -29,7 +29,7 @@ run:
     set +x
     if [[ -n ${GKE_PROJECT_ID} ]]; then
       # Delete load balancers
-      while read -r line; do
+      while read -r line && [[ -n ${line} ]]; do
           namespace=$(echo $line | awk '{print $1;}')
           svc_name=$(echo $line | awk '{print $2;}')
           attempts=0
@@ -49,13 +49,15 @@ run:
       done <<< $(kubectl get svc --all-namespaces | grep LoadBalancer)
 
       # Persistant storage info
-      export PERSISTANT_STORAGE=$(kubectl -n dispatch get pv -o json | jq -r '.items[0].spec.gcePersistentDisk.pdName')
+      export PERSISTANT_STORAGE=$(kubectl -n dispatch get pv -o json | jq -r '.items[0].spec.gcePersistentDisk.pdName // empty')
 
       # Delete cluster
       gcloud container clusters delete --quiet $(cat cluster/name)
 
       # Delete persistent storage
-      gcloud compute disks delete --quiet ${PERSISTANT_STORAGE}
+      if [[ -n ${PERSISTANT_STORAGE} ]]; then
+        gcloud compute disks delete --quiet ${PERSISTANT_STORAGE}
+      fi
     else
       helm delete --purge dispatch api-gateway openfaas ingress || true
       kubectl delete namespace kong dispatch openfaas || true
--- a/ci/e2e/cleanup.yml
+++ b/ci/e2e/cleanup.yml
@@ -15,6 +15,8 @@ inputs:
 params:
   GKE_KEY:
   GKE_PROJECT_ID:
+  MAX_ATTEMPTS: 18
+  TIMEOUT: 10
 
 run:
   path: /bin/bash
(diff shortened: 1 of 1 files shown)


kars7e:
let's see what e2e says

rjew:
Right now it only deletes the leftover resources that were created by the given cluster, but if you wanted to delete all leftover resources you could do something like this for volumes:
```
for disk in $(gcloud compute disks list --format=json | jq -r '.[] | .name'); do
    gcloud compute disks delete --quiet $disk
done
```
But that would delete all disks even ones that may still be in use. 

I don't think there is a simple way to determine if a resource is leftover and after deleting the cluster you won't be able to use kubectl to query the info needed anymore.

berndtj:
What is the status of this change?

kars7e:
I think @rjew's approach is good, I don't see any reliable strategy to clean up resources using gcloud without inspecting the cluster first. However, if we have access to cluster, I would rather use kubectl directly, using similar approach I posted in https://github.com/vmware/dispatch/pull/285#issuecomment-371364153. @rjew, could you make this change?

We still need to delete the orphaned resources we have today, and to be able to clean up resources that will slip through our regular clean-ups. I did some research, and I see some possible ways of doing that. They are not completely safe, e.g. if some resources are in creating/deleting state, they could  get deleted by these methods, so we need to run them when there is no CI job operating. Nevertheless, they should do the job. I see following ways:

** Disks **
* disks have `users` attribute which lists their attachments. Specifically, `gcloud compute disks list --filter "users=null"` returns disks without any attachments. Those are good candidates to be deleted.
* disk names include the cluster. `gcloud compute disks list --filter "name~dispatch-ci"` will return only disks for clusters that were created by CI.

Above, combined with `--format` flag and some `tr` will give this one-liner:
```bash
gcloud compute disks delete $(gcloud compute disks list --filter "name~dispatch-ci AND users=null" --format "value(name)" | tr '\n' ' ')
```

**  Load Balancers **
This is a bit more complicated, as a single Load Balancer is built on multiple sub-resources:
* a frontend (a forwarding rule)
* a backend (a target pool and health check)
But there is still hope. you can find orphaned backends by iterating over them and checking the health:
```bash
for tp in $(gcloud compute target-pools list --format "value(name)");  
do
    gcloud compute target-pools get-health $tp --format json;
done
```
The above will return `[]` for pools that do not have existing instances anymore. Those are good candidates for deletion. Before doing that, a forwarding rule matching it must be deleted:
forwarding rule that matches it:
```bash
gcloud compute forwarding-rules delete --region us-west1 $(gcloud compute forwarding-rules list --regions us-west1 --filter "target~$tp" --format "value(name)")
```
Then pool:
```bash
gcloud compute target-pools delete $tp
```
And finally, health check can be deleted:
```bash
gcloud compute http-health-checks delete $(gcloud compute target-pools describe $tp --format "value(healthChecks,0)")
```

rjew:
Sure I'll work on making the change you mentioned

kars7e:
Awesome! I will take care of that one-off cleanup script to delete currently orphaned resources.

rjew:
Changed the deletion of the load balancers from gcloud to kubectl. Slightly more complicated since load balancers are deleted asynchronously so we have to wait for the load balancer resources to actually be cleaned up before deleting the cluster.

I looked into deleting the persistent storage (pvc) using kubectl but there doesn't seem to be a reliable way to delete the pvc using kubectl.
1. Persistent storage cannot be deleted while the it is still in use. So one way I tried doing this was deleting the deployments related to the pvc but there doesn't seem to be a reliable way to determine if all the related resources have been deleted yet before deleting the pvc. The only event I could listen for is the deletion of the replica set but it still takes some time after that before we can delete the pvc.
2. After clearing the resources in use we can delete the pvc. But after the deleting the pvc we run into a similar problem in that there doesn't seem to be a way to determine if the storage has been deleted yet. It is deleted asynchronously like the load balancers but there is no event emitted that tells when it is successfully deleted. So if we delete the cluster too early, the kubernetes service will not have finished cleaning up the storage and it will become orphaned on google cloud.

It seems that to delete the persistent storage using only kubectl we would have to sleep an arbitrary amount of time after each deletion but that won't be able to reliably guarantee that the storage has been deleted.

kars7e:
This looks good, nice job with with the loadbalancers. There was one issue, `awk` is not available in the image. I will push new CI worker image

kars7e:
@rjew can you update the image in cleanup task to `vmware/dispatch-k8s-ci:v0.0.11` and rebase the change?

rjew:
@kars7e Updated image

kars7e:
@rjew the job failed: https://ci.dispatchframework.io/teams/main/pipelines/dispatch-e2e/jobs/tests-riff/builds/8 looks like yo need to handle the case when `kubectl get pv` returns empty list, which is possible because we're running `dispatch uninstall` now (so in successful run, it won't have PV leftovers).

kars7e:
Looks like service account is missing required permission to clean up disks. I fixed that and retriggered the job, let's see...

kars7e:
Merged, good job @rjew !

PR #292, opened 2018-03-09
Adds create/delete/get CLI support for policy

zimengyang:
* Add create/delete/get CLI support for policy
* TODO: update CLI

### DIFF
--- a/pkg/dispatchcli/cmd/create.go
+++ b/pkg/dispatchcli/cmd/create.go
@@ -19,6 +19,7 @@ import (
 	apiModels "github.com/vmware/dispatch/pkg/api-manager/gen/models"
 	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
 	functionModels "github.com/vmware/dispatch/pkg/function-manager/gen/models"
+	policyModels "github.com/vmware/dispatch/pkg/identity-manager/gen/models"
 	imageModels "github.com/vmware/dispatch/pkg/image-manager/gen/models"
 	secretModels "github.com/vmware/dispatch/pkg/secret-store/gen/models"
 	"github.com/vmware/dispatch/pkg/utils"
@@ -59,6 +60,7 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 		Images     []*imageModels.Image       `json:"images"`
 		Functions  []*functionModels.Function `json:"functions"`
 		Secrets    []*secretModels.Secret     `json:"secrets"`
+		Policies   []*policyModels.Policy     `json:"policies"`
 	}
 
 	o := output{}
@@ -125,6 +127,17 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 				return err
 			}
 			o.Secrets = append(o.Secrets, m)
+		case utils.PolicyKind:
+			m := &policyModels.Policy{}
+			err := yaml.Unmarshal(doc, &m)
+			if err != nil {
+				return errors.Wrapf(err, "Error decoding policy document &s", string(doc))
+			}
+			err = actionMap[docKind](m)
+			if err != nil {
+				return err
+			}
+			o.Policies = append(o.Policies, m)
 		default:
 			continue
 		}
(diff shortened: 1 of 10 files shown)


pkg/dispatchcli/cmd/create.go
```diff
@@ -158,6 +171,7 @@ func NewCmdCreate(out io.Writer, errOut io.Writer) *cobra.Command {
 				"BaseImage": CallCreateBaseImage,
 				"Function":  CallCreateFunction,
 				"Secret":    CallCreateSecret,
+				"Policy":    CallCreatePolicy,
```
neosab: While you are at it can you make this `utils.PolicyKind` and similar change to other keys?

pkg/dispatchcli/cmd/create_policy.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
+
+	policy "github.com/vmware/dispatch/pkg/identity-manager/gen/client/policy"
+	models "github.com/vmware/dispatch/pkg/identity-manager/gen/models"
+)
+
+var (
+	createPolicyLong = i18n.T(`Create a dispatch identity manager policy`)
```
neosab: `Create a dispatch policy` would suffice

pkg/dispatchcli/cmd/create_policy.go
```diff
@@ -0,0 +1,97 @@
...
+
+	createPolicyExample = i18n.T(`  dispatch create policy test_policy -s xyz@example.com -a get,update -r image,api,function`)
```
neosab: Whitespace in the beginning

neosab: Also, it will be nice to some proper examples like with long flags
```
# Create a policy by specifying a single subject, action, and resource
dispatch create policy example_policy --subject user1@example.com --action get --resource *

# Create a policy by specifying multiple subjects, actions, and resources
dispatch create policy example_policy --subject user1@example.com,user2@example.com --action get,create,delete,update --resource image,function,base-image,secret

````

File: pkg/dispatchcli/cmd/create_policy.go [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+		Long:    createPolicyLong,
+		Example: createPolicyExample,
+		Args:    cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			err := createPolicy(out, errOut, cmd, args)
+			CheckErr(err)
+		},
+	}
+
+	cmd.Flags().StringVarP(&subject, "subject", "s", "*", "subjects of policy rule, separated by comma")
```
neosab: 1. Let's not have a default `*`, same for below action & resource

2.Also, can we make this StringSliceP? I would like users to specify a flag multiple times in case it becomes harder to type with comma separated by a single flag. 
e.g 
`create policy --subject user1@example.com --subject user2@example.com ...`

pkg/dispatchcli/cmd/create_policy.go
```diff
@@ -0,0 +1,97 @@
...
+	*policyModel = *created.Payload
+	return nil
+}
+
+func createPolicy(out, errOut io.Writer, cmd *cobra.Command, args []string) error {
+
+	policyName := args[0]
+	policyRules := models.PolicyRules{
+		&models.Rule{
+			Subjects:  strings.Split(subject, ","),
```
neosab: If you use `StringSliceP` then you don't have to split this.

pkg/dispatchcli/cmd/delete_policy.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/spf13/cobra"
+	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
+	"golang.org/x/net/context"
+
+	policy "github.com/vmware/dispatch/pkg/identity-manager/gen/client/policy"
+	models "github.com/vmware/dispatch/pkg/identity-manager/gen/models"
+)
+
+var (
+	deletePolicyLong = i18n.T(`Delete a dispatch identity manager policy`)
```
neosab: `dispatch policy` should be enuf

pkg/dispatchcli/cmd/create.go
zimengyang: fixed, constants replaced with utils.*Kind. Include create.go and delete.go

pkg/dispatchcli/cmd/create_policy.go
zimengyang: fixed

pkg/dispatchcli/cmd/create_policy.go
zimengyang: An issue here is that when specify * as value, cmd will try to list all file in current directory. '*' has to be put inside quote marks. That's saying `--r "*"` instead of `--r *`. Is there a workaround?

pkg/dispatchcli/cmd/create_policy.go
zimengyang: fixed, change to StringSliceP()

pkg/dispatchcli/cmd/create_policy.go
zimengyang: fixed

PR #293, opened 2018-03-09
Add get API support for policy

zimengyang:
* add get API handler for policy
* TODO: add tests

### DIFF
--- a/pkg/identity-manager/handlers.go
+++ b/pkg/identity-manager/handlers.go
@@ -119,6 +119,7 @@ func (h *Handlers) ConfigureHandlers(api middleware.RoutableAPI) {
 	// Policy API Handlers
 	a.PolicyAddPolicyHandler = policyOperations.AddPolicyHandlerFunc(h.addPolicy)
 	a.PolicyGetPoliciesHandler = policyOperations.GetPoliciesHandlerFunc(h.getPolicies)
+	a.PolicyGetPolicyHandler = policyOperations.GetPolicyHandlerFunc(h.getPolicy)
 	a.PolicyDeletePolicyHandler = policyOperations.DeletePolicyHandlerFunc(h.deletePolicy)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/identity-manager/handlers.go
```diff
@@ -253,6 +254,26 @@ func (h *Handlers) getPolicies(params policyOperations.GetPoliciesParams, princi
...
+	defer trace.Tracef("get policy name '%s'", params.PolicyName)()
+	var policy Policy
+
+	name := params.PolicyName
+	if err := h.store.Get(IdentityManagerFlags.OrgID, name, entitystore.Options{}, &policy); err != nil {
+		log.Errorf("store error when getting policy '%s': %+v", name, err)
+		return policyOperations.NewGetPolicyNotFound().WithPayload(
+			&models.Error{
+				Code:    http.StatusNotFound,
+				Message: swag.String("policy not found when getting policy"),
```
neosab: Nit: `policy not found` would suffice

pkg/identity-manager/handlers.go
```diff
@@ -253,6 +254,26 @@ func (h *Handlers) getPolicies(params policyOperations.GetPoliciesParams, princi
 	return policyOperations.NewGetPoliciesOK().WithPayload(policyModels)
 }
 
+func (h *Handlers) getPolicy(params policyOperations.GetPolicyParams, principal interface{}) middleware.Responder {
+
+	defer trace.Tracef("get policy name '%s'", params.PolicyName)()
+	var policy Policy
+
+	name := params.PolicyName
+	if err := h.store.Get(IdentityManagerFlags.OrgID, name, entitystore.Options{}, &policy); err != nil {
```
neosab: Can you add FilterExists() to the entititystore.Options.Filter, even though we hard delete policies I would like to future proof. Also, can you make a similar change to deletePolicy?

pkg/identity-manager/handlers.go
zimengyang: fixed

pkg/identity-manager/handlers.go
zimengyang: fixed

Issue #103, opened 2018-01-10
ambiguity in function manager handler names: getFunctionRuns and getRuns


seanhuxy:
- [ ] Bug
- [ ] Feature
- [X] Enhancement

## Detailed Description

there're there handlers in function manager: 
getFunctionRuns,
getRuns,
getRun

their functions and required parameters are ambiguous, in fact:
getFunctionRuns --> get function runs by function name AND optionally run id
getRuns --> get all functions runs 
getRun --> get function run only by run id

It might be ok, but when it comes to dipatch cli get_runs.go there're
getFunctionRun --> get runs by function name AND run id
getFunctionRuns --> get runs by function name
getRuns --> get all runs 
which I cannot really tell the functions of them without carefully look into the code. 

## Possible Implementation

I suggest to they should renamed to something better reflect their real functions 
e.g. 
getRunsByFunctionName
getRuns
getRun

or maybe combine getRuns and getFunctionRuns if possible

## Complexity

- [X] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [X] Low - Annoyance, but does not impact business or functinality
- [ ] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

## Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

berndtj:
I think we could probably just get by with "getRuns" and "getRun", with everything else just being a filter on getRuns.  I agree that the current handlers are ambiguous.

PR #303, opened 2018-03-13
Fixes ambiguity in function manager handler names and REST API url conflict

### LINKED ISSUE #102: conflict in REST API url paths in function-manager
- [X] Bug
- [ ] Feature
- [ ] Enhancement

## Detailed Description

the url path of getRuns is /v1/function/runs now, 
it will be conflicted with /v1/function/{functionName} when there's a function called "runs", it may not happen but indeed a potential bugs.

## Possible Implementation

change the url path of getRuns to something like /v1/runs or /v1/functionruns or whatever won't cause the conflict.

## Complexity

- [X ] Low - Simple enhancement or bug fix, no architectural changes or refactoring

## Impact

- [X] Low - Annoyance, but does not impact business or functinality
- [ ] High - Blocker

## Your Environment

rjew:
Fixes #102 and #103 
* url path of getRuns was /v1/function/runs, is now /v1/runs
* url path of getRun was /v1/function/(function-name)/runs/(run-id), is now /v1/runs/(run-id)
* getFunctionRuns is now just a filter on getRuns
* getRuns now optionally accepts functionName as a filter
* getRun only depends on function id and can optionally include a function name as query (to satisfy CLI command where function name has to be specified)


swagger/function-manager.yaml
```diff
@@ -224,7 +189,7 @@ paths:
           description: Internal error
           schema:
             $ref: '#/definitions/Error'
-  /{functionName}/runs:
+  /function/{functionName}/runs:
```
berndtj: I think we should remove this path.  This is equivalent to `/runs/?functionName=myfunc`

rjew: Sure I can remove that endpoint since they're equivalent

rjew: Well actually since we send POST requests to that path `/function/{functionName}/runs` to run a function, would it be better to just leave it then?

berndtj: Ah.  Well, I think the correct way would be to send to /runs and put the function name in the post payload.

rjew: Sure I'll work on doing that

PR #304, opened 2018-03-13
riff driver: use kafka directly vs. via riff http gateway

imikushin:
FunctionExecution:
- .ID -> .FunctionID
- rm .Name
- add .RunID

riff driver:
- use kafka
- Close()
- configure to use kafka broker

e2e tests:
- update e2e/tests-riff
- wait until the run is READY when checking for `.output`


charts/dispatch/charts/function-manager/templates/config-map.yaml
```diff
@@ -21,7 +22,7 @@ data:
           "funcNamespace": "{{ .Values.faas.openfaas.namespace }}"
         },
         "riff": {
-          "gateway": "{{ .Values.faas.riff.gateway }}",
+          "kafkaBrokers": ["{{ index .Values.faas.riff.kafkaBrokers 0 }}"],
```
kars7e: helm uses [sprig](https://godoc.org/github.com/Masterminds/sprig), so this should work:
```
"kafkaBrokers": ["{{ join "," .Values.faas.riff.kafkaBrokers }}"],
```

pkg/utils/try-close.go
```diff
@@ -0,0 +1,14 @@
+///////////////////////////////////////////////////////////////////////
```
kars7e: NIT: should use _ instead of - in file name (but TBH, `close.go` is a perfectly fine name for this as well :) )

imikushin: yep, `close.go` works

charts/dispatch/charts/function-manager/templates/config-map.yaml
imikushin: Thanks! Fixed

PR #305, opened 2018-03-13
Add update CLI/API support for Policy

zimengyang:
* Add CLI and API support for updating Policy
* Change Policy output format
* Add tests for get/update handler.


zimengyang:
Tests for update/get handler added.

PR #307, opened 2018-03-13
Batch resource creation file path to function definition and CLI output

### LINKED ISSUE #295: Accept path to function from batch create/update
Creating functions via manifest file (`examples/seed.yaml`), the function needs to be entirely defined inline.  This is less than ideal.  We should also accept a file path to the function `@pathToFile.py`

- [ ] Bug
- [ ] Feature
- [x] Enhancement

## Possible Implementation

The client should follow the path which is identified by the following panther `@pathToFile.py` (similar to curl), and unpack the file.  Then the client sends the contents as part of the payload to the server.

## Complexity

- [x] Low - Simple enhancement or bug fix, no architectural changes or refactoring

## Impact

- [x] Low - Annoyance, but does not impact business or functionality
- [ ] High - Blocker

## Your Environment

rjew:
Fixes #295 and #288 

* Resource type and name are outputted when created via manifest file
* Function definition can be specified in manifest as a path to file with notation `@pathToFile.py`

### DIFF
--- a/examples/seed.yaml
+++ b/examples/seed.yaml
@@ -47,33 +47,7 @@ tags:
 ---
 kind: Function
 name: hello-py
-code: |
-  #!/usr/bin/env python
-  #######################################################################
-  ## Copyright (c) 2017 VMware, Inc. All Rights Reserved.
-  ## SPDX-License-Identifier: Apache-2.0
-  #######################################################################
-  """
-  Example function "Hello World"
-
-  ** REQUIREMENTS **
-
-  * image
-  dispatch create base-image python3 vmware/dispatch-openfaas-python-base:0.0.5-dev1 --language python3
-  dispatch create image python3 python3
-
-  Create a function:
-  dispatch create function python3 hello-python examples/python3/hello.py
-
-  Execute it:
-  dispatch exec hello-python --wait --input='{"name": "Jon", "place": "Winterfell"}'
-
-  """
-
-  def handle(ctx, payload):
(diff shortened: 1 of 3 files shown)


kars7e:
Super cool!

examples/seed.yaml
```diff
@@ -82,14 +56,7 @@ tags:
 ---
 kind: Function
 name: http-py
-code: |
```
tenczar: Could we leave one of these as the function 'literal' to show that we support both?

pkg/dispatchcli/cmd/create.go
```diff
@@ -105,17 +109,28 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
...
 			o.Images = append(o.Images, m)
+			fmt.Fprintf(out, "Created %s: %s\n", docKind, *m.Name)
 		case utils.FunctionKind:
 			m := &functionModels.Function{}
 			err = yaml.Unmarshal(doc, &m)
 			if err != nil {
 				return errors.Wrapf(err, "Error decoding function document %s", string(doc))
 			}
+			if strings.HasPrefix(*m.Code, "@") {
+				functionPath := (*m.Code)[1:]
```
tenczar: Shouldn't this be modified to work with the work-dir flag?

berndtj: Could you create a new issue?  Also, I'd like path supported for dependency files (e.g. requirements.txt/package.json)

PR #309, opened 2018-03-15
Version 0.1.7

pandom:
Updates version used


berndtj:
@pandom We updated the docs so that version no longer needs to be hard-coded.  Thanks for the PR though!

PR #310, opened 2018-03-15
Adding PowerCLI/NSX/VRA

pandom:
Adds image for PowerCLIcore with powernsx and powervra


kars7e:
As discussed, PowerCLI image can be created using following command:
```
dispatch create image powercli powershell-base --runtime-deps requirements.psd1
```

where `requirements.psd1` looks like this:
```
@{
 "VMware.PowerCLI" = "latest"
}
```

PR #311, opened 2018-03-15
Run Riff in e2e CI tests

kars7e:
**Changes**:
* Adds separate CI job to test with Riff. Runs in parallel with OpenFaas job.
* Unifies e2e tests, skipping those that cannot run on Riff. Set `FAAS` variable when running locally (defaults to openfaas)
* CI Makefile to trigger individual CI tasks. See comments in Makefile for details


kars7e:
Indeed, PKS folks are using templating engine to reduce the amount of it (yes, templates in templates...). ours are smaller so maybe we can cope with it for now, but something to think about in future.

PR #313, opened 2018-03-16
Removed path /function/{functionName}/runs

rjew:
Fixes issue referenced here https://github.com/vmware/dispatch/pull/303#pullrequestreview-103613668 

Path /function/{functionName}/runs is removed since it is equivalent to /runs?functionName={functionName}. runFunction (POST) and getFunctionRuns (GET) can now be done using that endpoint.


kars7e:
Note, because of GKE pulling out version 1.8.7 from their available k8s versions, we need to update in CI: https://github.com/vmware/dispatch/pull/314 must go first and this patch must be rebased.

kars7e:
#314 merged, I've updated the branch to retrigger the CI

kars7e:
@rjew you must manually rebase & push this branch, CI got stuck and won't retrigger automatically.

rjew:
There's an issue with the e2e test failing that has to do with api gateway. I've been looking into it and should have a fix up soon.

rjew:
Fixed the api errors and passes e2e test now.

Issue #269, opened 2018-03-05
Decouple language support from FaaS drivers


imikushin:
Add riff support for all languages currently supported by openfaas driver. 

- [ ] Bug
- [x] Feature
- [ ] Enhancement

## Context

We need to provide the same level of support for Riff and OpenFaaS. #256 is related. 

## Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [x] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [x] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

imikushin:
Adding language support could be trivial if FaaS-specific function adapter only needed to use one interface to call function implementations. All language-specific runtime code could be packaged in a single docker image. 


# Problem

Currently, it’s hard to add support for new languages to Dispatch: 

1. To add a new language, it should be added to all FaaSs currently supported in a FaaS-specific way. 
2. To add a support for a new FaaS to Dispatch, one not only needs to implement the FaaS driver, but also add the languages in a FaaS-specific way. As a consequence, language support differs between FaaS-drivers.
3. It’s impossible to add a language without rebuilding Dispatch.


# Proposed solution

Base-image should include everything required to support a language, including:

- The language runtime and its standard library — usually installed as a system package
- Function adapter — a simple HTTP server implementing Function Runtime API (see below)
- Dockerfile template to build the function image

At function image building phase, function-manager would extract the template files from the base-image and render them using the function source code and metadata. The resulting Dockerfile are used to build the function image. 


## Function Runtime HTTP API

It’s a low level function invocation API typically implemented by a simple HTTP server written in the language we’re building the base image for. This API is called by the FaaS driver (through the FaaS engine) at function invocation time: 

- OpenFaaS would call it through its faas-netes proxy
- Riff would call it through the function-sidecar

Endpoint: /

**Request**
Method: POST
Headers:

- Content-Type: application/json
- Accept: application/json

Body:

- context
- payload

**Response**
Headers:

- Content-Type: application/json

Body:

- context
  - error — error object/value, null if no error occured
  - logs — array of log strings from the function run
- payload


## Required changes

**Base Images**

1. Add the Function Runtime HTTP API server.
2. Add function image Dockerfile template. 

**Function-manager**

1. Remove built-in FaaS- and language-specific function image templates.
[...]

PR #320, opened 2018-03-20
Proposal: Decouple language support from FaaS drivers

imikushin:
The spec for #269

### DIFF
--- a/docs/_specs/lang-pack/lang-pack.md
+++ b/docs/_specs/lang-pack/lang-pack.md
@@ -20,7 +20,10 @@ Base-image should include everything required to support a language, including:
 
 - The language runtime and its standard library — usually installed as a system package
 - Function adapter — a simple HTTP server implementing Function Runtime API (see below)
+- Dockerfile template to install system and runtime dependency packages
 - Dockerfile template to build the function image
+
+At image build time, image-manager would use the Dockerfile template to install system and runtime dependencies.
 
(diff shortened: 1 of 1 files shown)


kars7e:
One more thing came to my mind. This proposal doesn't address the runtime dependencies feature. Today, when creating an image, a user can specify runtime dependencies, which are language-dependent. This is implemented in image manager, and if a new language is added, it would also require implementation there.

berndtj:
How will this work with non JSON payloads?

I think it's worth pointing out that I don't think the point of this is to allow the users/adminstrators of a dispatch installation to add language support, but it's more for us to add languages to existing installations by updating/adding to the base-images. At least that's how I read it.

@kars7e makes a good point, can we pull in the dependency related code into the image some how?  Otherwise we still need to rebuild when supporting a new language.

imikushin:
Thanks for comments! 

If the admin is adding a language runtime image (or "base-image"), they are expected to be able to install the Function Runtime API implementation for that language into the image. We're going to provide examples for that (for JS, Python and PowerShell - at least).

imikushin:
@kars7e I'll add the runtime dependency manifest support - great catch!

berndtj:
I understand this is to allow Dispatch more language flexibility.  I'm just stating that there is still some overhead to supporting a language (it's just all contained in the base image itself now).  This could of course come from a dispatch administrator, but would likely come from us.  Regardless we need to document it well!

imikushin:
This is now obsolete as #339 already implemented the feature. Closing in favor of upcoming documentation update.

Issue #282, opened 2018-03-07
Bootstrap mode should only allow policy requests


neosab:
Bootstrap mode should only allow policy requests and deny any other requests. This ensures no other operation can be performed in the system and in a way forces the operator to disable the bootstrap mode.

- [ ] Bug
- [ ] Feature
- [X] Enhancement

## Complexity

- [X] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [X] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

## Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

neosab:
To be clear:
1. The bootstrap user should be allowed access to only 'iam' resources.
2. Any other authenticated user should be presented with a 503 while in the bootstrap mode.

PR #323, opened 2018-03-20
Update auth for bootstrap user/mode

zimengyang:
* Remove bootstrap user info from secret when bootstrap mode disabled
* Only accept IAM resource operation of bootstrap user
* Add log info in identity manager for accept/forbidden details
* Add tests for bootstrap user/mode access cases


zimengyang:
should fix issue #306 and #282

PR #325, opened 2018-03-20
Topic/ntenczar/cli update image

tenczar:
Change adds batch support for image updates.


pkg/dispatchcli/cmd/update_image.go
```diff
@@ -0,0 +1,125 @@
...
+		img.RuntimeDependencies = &runtimeDependencies
+		changed = true
+	}
+
+	if !changed {
+		fmt.Fprintf(out, "No changes made")
+		return nil
+	}
+
+	CallUpdateImage(&img)
```
zimengyang: maybe might we check the status of CallUpdateImage() in case that the api call failed for some reason?

examples/seed.yaml
```diff
@@ -47,7 +47,7 @@ tags:
 ---
 kind: Function
 name: hello-py
-code: '@examples/python3/hello.py'
```
berndtj: If you change this back, the docs must also reflect this change.  We've caused a lot of headaches because of this change.

berndtj: Looks like the quickstart at least was never updated, so this change would be good.  But make sure we're not missing other places

pkg/dispatchcli/cmd/update_image.go
berndtj: Good call!

examples/seed.yaml
tenczar: Ok, I'll make sure to check.

pkg/dispatchcli/cmd/update_image.go
tenczar: I've removed this code because it was used to support the command line flag version of updating images. This function will be used in the future to support the file editor version of resource update.

PR #326, opened 2018-03-21
API gateway form encoding support

### LINKED ISSUE #294: Accept and form encoding from the API gateway
The API gateway should accept and transform form encoded data to JSON and pass it on to the function/event.  This makes building some web front-ends easier

- [ ] Bug
- [ ] Feature
- [x] Enhancement

## Complexity

## Impact

- [ ] High - Blocker

## Your Environment

rjew:
Fixes #294 
* API gateway now accepts content-types: json & form encoding
* Form encoded data is transformed to json before passing it to the function/event
* Updated tests to use content type header and added new tests to test content type functionality


charts/kong/dispatch-transformer/handler.lua
```diff
@@ -84,15 +89,34 @@ local function substitute_payload(conf, result)
...
-    result[conf.substitute.input] = json_data
+  local header = ngx.req.get_headers()
+  if is_json_body(header) then
+    local ok, json_data = parse_json(data)
+    if ok then
+      ngx.log(ngx.DEBUG, "request body is json")
+      result[conf.substitute.input] = json_data
+    else
+      ngx.log(ngx.DEBUG, "request body is not json")
+      result[conf.substitute.input] = data
```
berndtj: This seems wrong.  Shouldn't we return an HTTP error if the content type is application/json, but the payload doesn't match that?

berndtj: I guess that's just continuing existing behavior, but it seems wrong

charts/kong/dispatch-transformer/handler.lua
```diff
@@ -84,15 +89,34 @@ local function substitute_payload(conf, result)
...
+    end
+  elseif is_form_urlencoded_body(header) then
+    local form_urlencoded_data, err = ngx.req.get_post_args()
+    if not err then
+      ngx.log(ngx.DEBUG, "request body is x-www-form-urlencoded")
+      result[conf.substitute.input] = form_urlencoded_data
+    else
+      ngx.log(ngx.DEBUG, "request body is not x-www-form-urlencoded: " .. err)
```
berndtj: Again, this seems wrong.

charts/kong/dispatch-transformer/handler.lua
rjew: Yeah that's a good point. I was just extending the existing behavior. I'll look into throwing an HTTP error when the payload doesn't match the content type or when the content type is not supported or specified.

charts/kong/dispatch-transformer/handler.lua
kars7e: looks like error is returned below that? Or is this comment outdated?

rjew: Yeah that comment is outdated. I returned the error right below that line.

rjew:
So it seems to be failing the e2e test when creating a function with default secret. I tested it locally and it seems to be working fine. It's probably timing out too early before the function can finish creating since the later test that executes the given function passes.

PR #327, opened 2018-03-21
[cli/update] update e2e tests to use file based update mechanic

tenczar:
Small change to e2e tests to use the file based mechanic for updating resources. Also prunes the command line based update mechanism. This will be updated in the future to allow live editing at the command line.


e2e/tests/images.bats
```diff
@@ -74,14 +74,14 @@ load variables
...
-    run dispatch update base-image nodejs6-base --language python3
+@test "Update images" {
+    run dispatch update --work-dir ${BATS_TEST_DIRNAME} -f images_update.yaml
     assert_success
     run_with_retry "dispatch get base-image nodejs6-base --json | jq -r .language" "python3" 1 0
     assert_success
 
-    run dispatch update base-image not-found --language nodejs
-    assert_failure
+    #run_with_retry "dispatch get image python3 --json | jq -r .tags.0.key" "update" 1 0
```
berndtj: So what happens now when you update an image that doesn't exist?  That appears to be what the old test was doing.

e2e/tests/images_update.yaml
```diff
@@ -0,0 +1,8 @@
+kind: BaseImage
+name: nodejs6-base
+dockerUrl: vmware/dispatch-nodejs6-base:0.0.1-dev1
+language: python3
+tags:
+  - key: role
+    value: test
+---
```
berndtj: is this supposed to be here?

tenczar: I had added a test for images as well, but that was failing because the PR to support image updates hadn't been pushed yet. I'm currently adding that test back in and should have an updated version of this PR soon.

pkg/dispatchcli/cmd/update_api.go
```diff
@@ -17,8 +15,8 @@ import (
 )
 
 var (
-	updateAPILong    = "update an api based on json"
-	updateAPIExample = `dispatch update api my_api --path /new/path`
+	updateAPILong    = "update a name api "
```
berndtj: shouldn't this be "update an API by name"?  Also, are we being consistent with the phrasing across the CLI?

pkg/dispatchcli/cmd/update_api.go
```diff
@@ -17,8 +15,8 @@ import (
...
+	updateAPIExample = "dispatch update api my_api"
```
berndtj: Should the example include a path to a file?

pkg/dispatchcli/cmd/update_baseimage.go
```diff
@@ -19,22 +16,8 @@ import (
...
-		"reason": null,
-		"tags": [
-			{
-				"key": "role",
-				"value": "test"
-			}
-		]
-	}`
-	imageURL string
+	updateBaseImageLong    = "Updates a named base image in a file editor"
```
berndtj: Does the file editing work?

pkg/dispatchcli/cmd/update_baseimage.go
```diff
@@ -19,22 +16,8 @@ import (
...
+	updateBaseImageExample = "dispatch update base-image python3-base"
```
berndtj: Shouldn't this point to a file also?

pkg/dispatchcli/cmd/update_baseimage.go
```diff
@@ -66,37 +49,10 @@ func NewCmdUpdateBaseImage(out io.Writer, errOut io.Writer) *cobra.Command {
...
-		baseImage.Language = swag.String(language)
-	}
-
-	err = CallUpdateBaseImage(&baseImage)
-	if err != nil {
-		return err
-	}
-
-	fmt.Fprintf(out, "Updated base image: %s\n", baseImageName)
+	//TODO: Implement file writer update mechanism
```
berndtj: Should we just remove all of the update_*.go files?  Seems none of them currently work.  We are waiting for the editor support, correct?

tenczar: That's right. I can remove all of them and just move all the Call*Update methods into the update.go file.

PR #328, opened 2018-03-22
Update go to 1.10 and don't run coverage on every PR

kars7e:
* Update docker image to go1.10
* Update swagger gen files to latest
   *NOTE:* Small fixes were required to make services work with updated swagger gen files. Those revolve around the fact that go-swagger does not create a separate type if returned type is a slice anymore (it just uses the slice directly).
* Don't run coverage on every PR
* Fix a bug in Makefile which were causing BUILD to be evaluated every time it was referenced, instead of once for the entire Makefile


kars7e:
Looks like there is a race condition in GKE cleanup now, volume is scheduled for deletion but is still returned by kubectl, then when gcloud disk delete is attempted, disk is already gone. I will post a separate patch. Retrigger was enough to clear CI

PR #338, opened 2018-03-27
[cli/update] add update functionality for functions in dispatch cli

tenczar:
Adds function update to dispatch cli. Note this causes the creation of a new function in the FaaS.


pkg/function-manager/handlers.go
```diff
@@ -456,6 +454,8 @@ func (h *Handlers) updateFunction(params fnstore.UpdateFunctionParams, principal
 		})
 	}
 
+	id := uuid.NewV4().String()
+	e.BaseEntity.ID = id
```
kars7e: This does not seem right, why do we do that?

tenczar: This was to force the creation of a new function in the FaaS. This id is not used internally, but it is used as a key for creating new functions.

kars7e: Can you add a comment to this in the code? I would never guess it myself

berndtj: Wait, this is not a good idea.  We should not be changing the ID of our entities.  The correct way to handle this is to create a new field called `FaasID` or `FunctionID` which maps to the FaaS generated ID/name.  We should not change the ID of our entities, it isn't an "update" anymore.

tenczar: So, what you have described for FaasID is exactly how we are using ID. ID is not used internally by dispatch. We uniquely identify functions by a key which is made up of the type of the entity and the name. I can easily add this extra field if you feel this is confusing as implemented.

berndtj: ID is supposed to represent the key of an entity which is unique across time (unlike name).  Even though we don't expose it directly today, I don't want us to change the meaning.  An entity with a different ID is a NEW entity, not an updated one.  I expect the dispatch function entity to continue to have the same ID across updates, even if the FaaS function does not.

pkg/function-manager/handlers.go
```diff
@@ -456,6 +454,9 @@ func (h *Handlers) updateFunction(params fnstore.UpdateFunctionParams, principal
 		})
 	}
 
+	// generating a new UUID will force the creation of a new function in the underlying FaaS
+	id := uuid.NewV4().String()
+	e.BaseEntity.ID = id
```
berndtj: I thought I already commented on this, but we shouldn't change the ID of an entity (they should be considered immutable).  What we should do is add another property like "FaaSID" which is used as the name for the function within the FaaS.

tenczar: You definitely did make this comment. This is what I have been working to fix. I'm currently stuck trying to figure out why the deployment in OpenFaaS is never ready even though the pod seems to indicate that it is 'serving'. I will ping Ivan.

PR #339, opened 2018-03-27
WIP: Pack language support into base images

imikushin:
In line with proposal in #269.

Also make a few supporting changes: 

Make image building more robust: Docker API Build command returns
200 OK and streams its progress, including errors.

Do not suppress docker API output when building images.

Disable openfaas health check: it checks for /tmp/.lock file which does
not tell you if the function is ready.


images/base-images/nodejs/function-server/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "0.0.1",
+  "description": "Node.js Dispatch function server",
+  "private": true,
+  "main": "server.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "dependencies": {
+    "body-parser": "^1.18.2",
+    "express": "^4.15.5"
```
berndtj: Is there any danger of user defined dependencies conflicting?

pkg/functions/riff/driver_test.go
```diff
@@ -102,8 +102,7 @@ func TestDriver_Create(t *testing.T) {
 	}
 
 	err := d.Create(&f, &functions.Exec{
-		Image:    "imikushin/dispatch-riff-nodejs6-base:0.0.3-dev1",
-		Language: "nodejs6",
+		Image: "imikushin/dispatch-riff-nodejs6-base:0.0.3-dev1",
```
berndtj: We shouldn't have references to external images.  I can create a repo for you if you need.

pkg/image-manager/entities.go
```diff
@@ -68,15 +45,13 @@ type SystemDependencies struct {
 
 // RuntimeDependencies defines a runtime dependency type
 type RuntimeDependencies struct {
-	Format   string `json:"format"`
 	Manifest string `json:"manifest"`
```
berndtj: Why remove format?

pkg/image-manager/entities.go
```diff
@@ -68,15 +45,13 @@ type SystemDependencies struct {
...
 }
 
 // Image defines an image type
 type Image struct {
 	entitystore.BaseEntity
 	DockerURL           string              `json:"dockerUrl"`
-	Language            Language            `json:"language"`
```
berndtj: It's not bad to keep this, as it's effectively caching the value so you don't need to look it up on the base image.

pkg/image-manager/handlers.go
```diff
@@ -65,7 +65,6 @@ func baseImageEntityToModel(e *BaseImage) *models.BaseImage {
 	m := models.BaseImage{
 		CreatedTime: e.CreatedTime.Unix(),
 		DockerURL:   swag.String(e.DockerURL),
-		Language:    models.Language(e.Language),
```
berndtj: I'm confused.  How are we conveying the language of the [base]image back to the user?

images/base-images/nodejs/image-template/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+ARG SYSTEM_PACKAGES_FILE=system-packages.txt
+COPY ${SYSTEM_PACKAGES_FILE} .
+
+RUN packages=$(cat ${SYSTEM_PACKAGES_FILE} | sed 's/ /-/' | uniq | paste -d' ' -); \
```
berndtj: Are you changing the format of the packages file?  It was a structured json document, but it looks like you're moving to a txt list.  I think we should keep the structure.  Also, I don't know why you are substituting?

images/base-images/python3/function-server/main.py
```diff
@@ -0,0 +1,55 @@
+import json
+import signal
+import sys
+import io
+
+import falcon
```
berndtj: This is fine... but we should look at just using the std lib.

pkg/functions/builder.go
```diff
@@ -51,78 +72,40 @@ func (ib *DockerImageBuilder) BuildImage(faas, fnID string, exec *Exec) (string,
...
+	err = images.BuildAndPushFromDir(ib.docker, tmpDir, name, ib.registryAuth, buildArgs)
 	return name, err
 }
 
-func cleanup(tmpDir string) {
-	defer trace.Tracef("rm dir: %s", tmpDir)()
-	os.RemoveAll(tmpDir)
+const functionFile = "function.txt"
+
+func ptr(s string) *string {
```
berndtj: or swag.String(...) :)

pkg/functions/builder_test.go
```diff
@@ -128,3 +91,23 @@ func errorMock(statusCode int, message string) func(req *http.Request) (*http.Re
 		}, nil
 	}
 }
+
+func Test_copyFunctionTemplate(t *testing.T) {
```
berndtj: We should really be consistent about our test naming.  Seems that you prefer the Test_ nomenclature, but everyone else is just using camel case.  We need to pick one.

imikushin: 👍 for consistency.

If there you want to write unit tests for `FooBar` and `fooBar` functions, how would you suggest tests be named? I'd name them `TestFooBar` and `Test_fooBar`.

pkg/functions/builder.go
imikushin: Sure! Although, in Go, reuse seems to be an anti-pattern :)

pkg/image-manager/entities.go
imikushin: It's not needed anymore

imikushin: The image building Dockerfile deals with the manifest natively

pkg/image-manager/entities.go
imikushin: Keeping language as informative field in the latest revision of this PR

images/base-images/nodejs6/Dockerfile
```diff
@@ -1,12 +1,22 @@
-FROM vmware/photon:2.0
+FROM vmware/photon2:20180302
 
-RUN tdnf install -y nodejs-7.7.4-4.ph2
+RUN tdnf install -y nodejs-8.3.0-1.ph2
```
kars7e: Should we call this image nodejs8-base instead? or just drop the version number. this "nodejs6" thing is kind of misleading now.

pkg/functions/riff/driver_test.go
kars7e: Do we still need this image? IIUC, we don't need FaaS-specific images anymore.

PR #342, opened 2018-03-27
Pass HTTP context into function context and fix API gateway bug

### LINKED ISSUE #187: pass http method thru function context
currently, there's no way for a function to know the original http method. While most applications APIs distinguish their CRUD operations from GET/POST/PATCH/DELETE http method. It will be great if we can also support that. 

- [ ] Bug
- [ ] Feature
- [ ] Enhancement

## Detailed Description

It requires to update api-gateway plugins, as well as some effects from function manage (i.e. retrieves the additional http headers like "origin-http-method" from the request and pass it to function context

## Complexity

## Impact

- [ ] Low - Annoyance, but does not impact business or functinality
- [X] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

## Your Environment

rjew:
Fixes #187

* HTTP context (path, method, headers, etc.) are passed into the function context through the api gateway
* HTTPContext field added to run
* Fixed bug with API gateway where the functionName query param would be applied globally, is instead applied on a per-api basis now


swagger/function-manager.yaml
```diff
@@ -421,6 +421,10 @@ definitions:
         type: array
         items:
           type: string
+      httpContext:
```
kars7e: seems like this attribute should be `readOnly`, as it cannot be manually set when executing a function through API.

kars7e:
Does it have to be passed using header? Could we instead pass it in the run payload body?

swagger/function-manager.yaml
rjew: Yeah it should probably be `readOnly`. I'll make the change.

rjew:
It could be passed through the payload body but wouldn't it be better for it to be stored in the function context similar to how we pass secrets since it isn't passed to the function directly by the user?

rjew:
Sure I'll work on adding some e2e tests.

kars7e:
What I meant is couldn't it be passed in the function manager API in the body of the call to run API. You used secrets as an example, and they are passed this way (in the body of the API call, not header). This would save you from parsing the header.

berndtj:
@kars7e This may have been on my bad advice.  I didn't realize that we already had a JSON envelope.  If we do, then there is no sense in using a header.  The header was to avoid having to modify the payload.

neosab:
Also, HTTP HEADERS have a size limit, on nginx I believe this is 8k. So not a good idea to pass these as HEADERS. http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers

rjew:
Added e2e test and updated it to being passed in the body instead of the header.

charts/kong/dispatch-transformer/handler.lua
```diff
@@ -198,6 +198,42 @@ local function insert_header_to_payload(conf, result)
...
+    -- Need to extract method out of request since always changed to POST
+    if not http_context["method"] then
+      local method = string.match(ngx.var.request, "%a+")
+      if method then
+        http_context["method"] = method
+      end
+    end
+
+    result[conf.substitute.http_context] = http_context
+    ngx.log(ngx.DEBUG, "after insert_http_context: payload: " .. cjson.encode(result))
```
tenczar: Could this potentially log sensitive information like cookies?

rjew: Good catch. It could potentially log out cookies. I'll edit it out.

PR #344, opened 2018-03-29
Bump k8s client to k8s-1.9.x

imikushin:
Vendor riff dependencies separately from the rest of the project as it does not compile against the new k8s.io/client-go version.

Encapsulate riff dependencies in `lib/riff/vendor`: `lib/riff` does not depend on any dispatch or
third party packages except `github.com/pkg/errors` and `github.com/sirupsen/logrus`.


Gopkg.lock
```diff
@@ -797,10 +784,12 @@
   revision = "eb3733d160e74a9c7e442f435eb3bea458e1d19f"
 
 [[projects]]
-  branch = "release-1.8"
```
berndtj: FYI, I'll be bumping these to `version = "kubernetes-1.9.1"`

berndtj:
I don't have a problem with this change, let's just get the tests to pass.

Also, in I've confirmed with @markpeek that we should move the faas drivers to their own repositories under github.com/dispatchframework.

Gopkg.lock
imikushin: Why not the latest stable in the 1.9.x branch?

berndtj: Isn't release more "stable" than HEAD on a branch? I'm not sure that I care, just that the client-go package is using branch = "release-6.0" which isn't clear which k8s version it supports.  Tagging based on release just seems more straight forward and definitive.

imikushin:
Factoring out faas drivers will be the next step

PR #346, opened 2018-03-29
Service manager initial commit

berndtj:
This is an initial commit.  It's still a work in progress, but it shouldn't break anything, and due to all the vendor changes, I'd like to commit this ASAP.  This change includes a working service manager which calls the kubernetes service catalog to provision and bind services.  I have only run/tested this manually.  There is still quite a bit of remaining work:
* charts/packaging
  - service catalog will likely remain a manual step in the short term
  - service manager may also remain disabled by default
* tests (hardly any unittests are written yet)
* CLI support
* e2e tests

This is a start though and can be demonstrated against the `UPS` example service broker from the kubernetes service catalog examples.


Gopkg.lock.old
```diff
@@ -0,0 +1,554 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
kars7e: Did you check this file in by accident?

berndtj: why yes, I did :).  Good catch

pkg/service-manager/clients/k8s_service_catalog.go
```diff
@@ -0,0 +1,436 @@
...
+func (c *k8sServiceCatalogClient) ListServiceInstances() ([]entitystore.Entity, error) {
+	instances, err := c.sdk.RetrieveInstances(c.config.CatalogNamespace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error retreiving service instance list")
+	}
+	var serviceInstances []entitystore.Entity
+	for _, instance := range instances.Items {
+
+		parameters := make(map[string]interface{})
+		err := json.Unmarshal(instance.Spec.Parameters.Raw, &parameters)
```
neosab: Similar to `ServiceBinding` looks like the `parameters` are optional. So may be check if it's defined :-
https://github.com/kubernetes-incubator/service-catalog/blob/4a138279dbf7bbb3b3d0e51dcca50858e6738604/pkg/apis/servicecatalog/types.go#L599

pkg/service-manager/clients/k8s_service_catalog.go
```diff
@@ -0,0 +1,436 @@
...
+			BaseEntity: entitystore.BaseEntity{
+				ID:     instance.ObjectMeta.Name,
+				Status: entitystore.StatusUNKNOWN,
+			},
+			ServiceClass: instance.Spec.ClusterServiceClassExternalName,
+			ServicePlan:  instance.Spec.ClusterServicePlanExternalName,
+			Parameters:   parameters,
+		}
+		for _, cond := range instance.Status.Conditions {
+			if cond.Type == v1beta1.ServiceInstanceConditionReady && cond.Status == v1beta1.ConditionTrue {
```
neosab: We should also handle the case when `cond.Type == v1beta1.ServiceInstanceConditionReady && cond.Status == v1beta1.ConditionFalse`? 

https://github.com/kubernetes-incubator/service-catalog/blob/f7c622f6491fc9155871dcd75cec4c4df006a418/pkg/controller/controller_instance.go#L996

pkg/service-manager/entities/entities.go
```diff
@@ -0,0 +1,221 @@
...
+	var tags []*models.Tag
+	for k, v := range e.Tags {
+		tags = append(tags, &models.Tag{Key: k, Value: v})
+	}
+
+	m := models.ServiceInstance{
+		CreatedTime:      e.CreatedTime.Unix(),
+		ID:               strfmt.UUID(e.ID),
+		Name:             swag.String(e.Name),
+		Kind:             utils.ServiceClassKind,
```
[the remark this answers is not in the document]
berndtj: Nice catch!

pkg/service-manager/clients/k8s_service_catalog.go
berndtj: Yeah, I need to verify some of these conditions through unit tests too.  Another good catch.

pkg/service-manager/clients/k8s_service_catalog.go
berndtj: Wow, you are thorough!  Also, thanks!

berndtj:
Great review @neosab!

pkg/service-manager/entities/entities.go
```diff
@@ -0,0 +1,221 @@
...
+	Bind             bool            `json:"bind"`
+	Binding          *ServiceBinding `json:"binding,omitempty"`
+}
+
+var statusMap = map[models.Status]entitystore.Status{
+	models.StatusCREATING:    entitystore.StatusCREATING,
+	models.StatusDELETED:     entitystore.StatusDELETED,
+	models.StatusERROR:       entitystore.StatusERROR,
+	models.StatusINITIALIZED: entitystore.StatusINITIALIZED,
+	models.StatusREADY:       entitystore.StatusREADY,
```
tenczar: Does it make sense to have updating?

pkg/service-manager/entities/entities_test.go
```diff
@@ -0,0 +1,24 @@
...
+///////////////////////////////////////////////////////////////////////
+
+package entities
+
+import (
+	"testing"
+)
+
+func TestServiceClassEntityToModel(t *testing.T) {
+
```
tenczar: Do we want some TODO's here? That would hopefully allow us to better track that these need to be fleshed out.

pkg/service-manager/handlers.go
```diff
@@ -0,0 +1,281 @@
...
+	a.ServiceInstanceAddServiceInstanceHandler = serviceinstance.AddServiceInstanceHandlerFunc(h.addServiceInstance)
+	a.ServiceInstanceGetServiceInstanceByNameHandler = serviceinstance.GetServiceInstanceByNameHandlerFunc(h.getServiceInstanceByName)
+	a.ServiceInstanceGetServiceInstancesHandler = serviceinstance.GetServiceInstancesHandlerFunc(h.getServiceInstances)
+	a.ServiceInstanceDeleteServiceInstanceByNameHandler = serviceinstance.DeleteServiceInstanceByNameHandlerFunc(h.deleteServiceInstanceByName)
+}
+
+func (h *Handlers) getServiceClassByName(params serviceclass.GetServiceClassByNameParams, principal interface{}) middleware.Responder {
+	defer trace.Trace("")()
+	e := entities.ServiceClass{}
+	err := h.Store.Get(flags.ServiceManagerFlags.OrgID, params.ServiceClassName, entitystore.Options{}, &e)
```
tenczar: Why not use the new find method to distinguish between not found and internal error?

pkg/service-manager/handlers.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (h *Handlers) getServiceInstanceByName(params serviceinstance.GetServiceInstanceByNameParams, principal interface{}) middleware.Responder {
+	defer trace.Trace("")()
+	si := entities.ServiceInstance{}
+
+	var err error
+	opts := entitystore.Options{
+		Filter: entitystore.FilterExists(),
+	}
+	err = h.Store.Get(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, opts, &si)
```
tenczar: Why not use the Find method?

pkg/service-manager/handlers.go
```diff
@@ -0,0 +1,281 @@
...
+		log.Warnf("Received GET for non-existent service instance %s", params.ServiceInstanceName)
+		log.Debugf("store error when getting service instance: %+v", err)
+		return serviceinstance.NewGetServiceInstanceByNameNotFound().WithPayload(
+			&models.Error{
+				Code:    http.StatusNotFound,
+				Message: swag.String(fmt.Sprintf("service instance %s not found", params.ServiceInstanceName)),
+			})
+	}
+	b := entities.ServiceBinding{}
+	err = h.Store.Get(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, opts, &b)
```
tenczar: why not use Find?

pkg/service-manager/handlers.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (h *Handlers) deleteServiceInstanceByName(params serviceinstance.DeleteServiceInstanceByNameParams, principal interface{}) middleware.Responder {
+	defer trace.Trace("")()
+	e := entities.ServiceInstance{}
+
+	var err error
+	opts := entitystore.Options{
+		Filter: entitystore.FilterExists(),
+	}
+	err = h.Store.Get(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, opts, &e)
```
tenczar: Why not Find?

pkg/service-manager/handlers.go
```diff
@@ -0,0 +1,281 @@
...
+			&models.Error{
+				Code:    http.StatusNotFound,
+				Message: swag.String("service instance not found"),
+			})
+	}
+	err = h.Store.Delete(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, &entities.ServiceInstance{})
+	if err != nil {
+		return serviceinstance.NewDeleteServiceInstanceByNameNotFound().WithPayload(
+			&models.Error{
+				Code:    http.StatusNotFound,
```
tenczar: Would this really be a not found situation? I imagine this would only happen as a race condition of trying to delete the same instance twice concurrently

swagger/service-manager.yaml
```diff
@@ -0,0 +1,374 @@
...
+      tags:
+      - serviceClass
+      summary: List all existing service classes
+      operationId: getServiceClasses
+      produces:
+      - application/json
+      parameters:
+      - in: query
+        name: broker
+        description: Broker name
```
tenczar: is this being used?

swagger/service-manager.yaml
```diff
@@ -0,0 +1,374 @@
...
+      tags:
+      - serviceInstance
+      summary: Get all service instances
+      description: List all service instances
+      operationId: getServiceInstances
+      produces:
+      - application/json
+      parameters:
+      - in: query
+        name: serviceclass
```
tenczar: Is this being used?

PR #351, opened 2018-04-04
Add Kafka backend for events and make it default

### LINKED ISSUE #255: Enhance event manager with Kafka support
- [ ] Bug
- [ ] Feature
- [x] Enhancement

## Detailed Description
Provide alternative event transport in form of Kafka Broker. It could be the default transport when using Riff as FaaS backend so that we reuse the same Kafka instance.

## Possible Implementation
Requires `events/transport` Kafka implementation, and Kafka chart to be installed.

## Complexity

## Impact

- [x] Low - Annoyance, but does not impact business or functionality
- [ ] High - Blocker

kars7e:
**Changes**:
* Adds new event transport based on Kafka Broker
* Kafka is now the default transport regardless of FaaS used. RabbitMQ can still be enabled via config flag.
* Kafka is now deployed to dispatch namespace instead of separate one, and an external deployment can be used (before it was hardcoded to use the built-in one).
* Fixes small issue in CI cleanup task, that was causing failure when deleting orphaned k8s services
* Also includes updated Gopkg.lock caused by the latest `dep`, which lists packages one per line.

**Pending Issues**
Kafka does not support wildcard topic matching. Because of that, subscriptions are now 1:1 (source and event types must be set without `*`). I will create a separate issue to discuss wildcard subscriptions (as it's related to multi-tenancy, performance, and other aspects).

Fixes #255

### DIFF
--- a/pkg/dispatchcli/cmd/install_config.go
+++ b/pkg/dispatchcli/cmd/install_config.go
@@ -54,10 +54,14 @@ openfaas:
 kafka:
   chart:
     chart: kafka
-    namespace: riff-system
+    namespace: dispatch
     release: transport
     repo: https://riff-charts.storage.googleapis.com
     version: 0.0.1
+  brokers:
+  - transport-kafka.dispatch:9092
+  zookeeperNodes:
+  - transport-zookeeper.dispatch:2181
 rabbitmq:
   chart:
     chart: rabbitmq
@@ -101,6 +105,7 @@ dispatch:
   bootstrapUser:
   insecure: false
   faas: openfaas
+  eventTransport: kafka
   #imageRegistry:
   #  name:
   #  username:
--- a/pkg/events/transport/kafka.go
+++ b/pkg/events/transport/kafka.go
@@ -5,4 +5,198 @@
 
 package transport
 
-// NO TESTS
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/Shopify/sarama"
+	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/vmware/dispatch/pkg/events"
+	"github.com/vmware/dispatch/pkg/trace"
+)
+
+// Kafka Implements transport interface using Kafka broker.
+type Kafka struct {
+	producer     sarama.SyncProducer
+	consumer     sarama.Consumer
+	producerOnly bool
+}
+
+// OptKafkaSendOnly creates producer only. Subscribe operation will panic
+func OptKafkaSendOnly() func(k *Kafka) error {
+	return func(k *Kafka) error {
+		k.producerOnly = true
+		return nil
+	}
+}
+
+// NewKafka creates an instance of transport based on Kafka broker.
+func NewKafka(brokerAddrs []string, options ...func(k *Kafka) error) (*Kafka, error) {
+	syncProducer, err := sarama.NewSyncProducer(brokerAddrs, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	k := Kafka{
+		producer: syncProducer,
+	}
+	for _, option := range options {
+		// TODO: handle errors from options
+		option(&k)
+	}
+	if k.producerOnly {
+		return &k, nil
+	}
+
+	k.consumer, err = sarama.NewConsumer(brokerAddrs, nil)
+	return &k, err
+}
+
+// Publish publishes an event
+func (k *Kafka) Publish(ctx context.Context, event *events.CloudEvent, topic string, tenant string) error {
+	defer trace.Tracef("topic: %s", event.EventType)()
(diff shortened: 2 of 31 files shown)


kars7e:
Added #352 to track the issue of wildcard subscriptions

pkg/dispatchcli/cmd/install_config.go
```diff
@@ -54,10 +54,14 @@ openfaas:
 kafka:
   chart:
     chart: kafka
-    namespace: riff-system
+    namespace: dispatch
```
berndtj: Should we just default kong and postgres to also deploy into the dispatch namespace?

pkg/events/transport/kafka.go
```diff
@@ -5,4 +5,198 @@
...
+	}
+
+	k.consumer, err = sarama.NewConsumer(brokerAddrs, nil)
+	return &k, err
+}
+
+// Publish publishes an event
+func (k *Kafka) Publish(ctx context.Context, event *events.CloudEvent, topic string, tenant string) error {
+	defer trace.Tracef("topic: %s", event.EventType)()
+	sp, _ := opentracing.StartSpanFromContext(
```
berndtj: Nice.  We need to do this everywhere.  Can we just combine our trace function (directly above) with this?  Obviously add some args to Tracef.

pkg/events/transport/kafka.go
```diff
@@ -5,4 +5,198 @@
...
+	)
+	defer sp.Finish()
+
+	msg, err := fromEvent(event)
+	if err != nil {
+		return errors.Wrapf(err, "error when creating Kafka message from CloudEvent for topic %s", topic)
+	}
+	msg.Topic = topic
+
+	err = injectSpan(sp, msg)
```
berndtj: Maybe, it's not so simple :)

pkg/events/transport/kafka.go
```diff
@@ -5,4 +5,198 @@
...
+	defer trace.Tracef("topic: %s", topic)()
+	sp, _ := opentracing.StartSpanFromContext(
+		ctx,
+		"Kafka.Subscribe",
+	)
+	defer sp.Finish()
+
+	doneChan := make(chan struct{})
+
+	// create partition consumer. Since we are creating only one consumer, we should messages from all partitions
```
berndtj: I think you're missing a word

tenczar: we should do what with the messages?

imikushin:
Looks solid. Great work!

pkg/events/transport/kafka.go
```diff
@@ -5,4 +5,198 @@
...
+			}
+		}
+	}()
+
+	return &subscription{done: doneChan}, nil
+}
+
+// Close closes the transport
+func (k *Kafka) Close() {
+	err := k.producer.Close()
```
tenczar: Do you need to close the consumer as well?

pkg/dispatchcli/cmd/install_config.go
kars7e: Yes, I think it's time to make the single namespace the default deployment model, I'll open an issue.

pkg/events/transport/kafka.go
kars7e: We could do that, however, it would not automatically retrieve the correlation IDs, I'm not sure how would that be represented in Jaeger. Worth a try.

pkg/events/transport/kafka.go
```diff
@@ -5,4 +5,198 @@
...
+
+// Kafka Implements transport interface using Kafka broker.
+type Kafka struct {
+	producer     sarama.SyncProducer
+	consumer     sarama.Consumer
+	producerOnly bool
+}
+
+// OptKafkaSendOnly creates producer only. Subscribe operation will panic
+func OptKafkaSendOnly() func(k *Kafka) error {
```
tenczar: Does this need to be a function that returns a function? OptKafkaSendOnly() doesn't do anything interesting in and of itself.

pkg/events/transport/kafka.go
kars7e: Yeah you need some correlation data carrier, within the code that can be `context.Context`, you need some extra code to serialize that when sending over the wire.

pkg/events/transport/kafka.go
kars7e: I think you might be right :) Will fix in a follow-up patch.

kars7e: Yep, Berndt already pointed that out... will fix in a follow-up patch!

pkg/events/transport/kafka.go
kars7e: good catch, will fix that too

pkg/events/transport/kafka.go
kars7e: I'm using functional options pattern: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

PR #354, opened 2018-04-04
CLI support for services manager

berndtj:
This commit adds CLI support for the services manager.  Additionally,
the following other changes are included:

* charts/installation for service manager
* fix deletion for service instances and bindings
* add soft-delete to entity store (other controllers need to start using this also)
* add service support to funciton manager
  * add --service option to function manager CLI/API
  * inject service bindings/secrets at function execution time


pkg/functions/secretinjector/secretinjector.go
```diff
@@ -8,23 +8,31 @@ package secretinjector
...
 type secretInjector struct {
-	client *secretclient.SecretStore
+	secretClient  *secretclient.SecretStore
+	serviceClient *serviceclient.ServiceManager
 }
 
 // New create a new secret injector
-func New(secretClient *secretclient.SecretStore) functions.SecretInjector {
-	return &secretInjector{client: secretClient}
+func New(secretClient *secretclient.SecretStore, serviceClient *serviceclient.ServiceManager) functions.SecretInjector {
```
kars7e: Could we define a new injector for service bindings here? This way secretinjector can remain as is, and we will have better separation of concerns.

pkg/dispatchcli/cmd/create_function.go
```diff
@@ -28,6 +28,7 @@ var (
 	schemaInFile          = ""
 	schemaOutFile         = ""
 	fnSecrets             = []string{}
+	fnServices            = []string{}
```
kars7e: You are defining this to be an empty slice, meaning if a user does not provide anything, it will still be sent do API as `[]`, not as nil. Later in API code you are checking if `services == nil` (which should be done in case someone is using API directly). That's ok in general, but it caused problems before when empty slice defined in subscription CLI would overwrite non-empty secrets defined for function. Might be worth checking if the empty slice is really what you meant here.

pkg/function-manager/handlers.go
```diff
@@ -139,6 +143,12 @@ func runModelToEntity(m *models.Run, f *functions.Function) *functions.FnRun {
 	} else {
 		secrets = append(secrets, m.Secrets...)
 	}
+	services := f.Services
+	if services == nil {
```
kars7e: Checking if `services == nil` when setting empty slice in CLI (see my other comment).

pkg/service-manager/handlers.go
```diff
@@ -249,33 +249,51 @@ func (h *Handlers) getServiceInstances(params serviceinstance.GetServiceInstance
...
+	err = h.Store.Get(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, opts, &i)
 	if err != nil {
 		return serviceinstance.NewDeleteServiceInstanceByNameNotFound().WithPayload(
 			&models.Error{
 				Code:    http.StatusNotFound,
 				Message: swag.String("service instance not found"),
 			})
 	}
-	err = h.Store.Delete(flags.ServiceManagerFlags.OrgID, params.ServiceInstanceName, &entities.ServiceInstance{})
+	err = h.Store.SoftDelete(&b)
```
kars7e: what's the background for this change? Why SoftDelete instead of Delete?

pkg/functions/secretinjector/secretinjector.go
berndtj: I was going to go that route... but I need to call the secrets API from the services injector also.  I can still make this change if you like.

pkg/dispatchcli/cmd/create_function.go
berndtj: ok.  I was more or less just copying the secrets impl.  I'll check.

pkg/service-manager/handlers.go
berndtj: Because, I can't actually delete the service instance until the binding is removed.  I get an error if I do.  Therefore, I need the entity marked for delete until I can actually delete it.  Also, this is good practice in general as if the entity is lost (it only exists on the channel), the actual deletion work will not get done.  We'll have just orphaned the resource.

pkg/dispatchcli/cmd/create.go
```diff
@@ -155,6 +157,18 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 			}
 			o.Policies = append(o.Policies, m)
 			fmt.Fprintf(out, "Created %s: %s\n", docKind, *m.Name)
+		case utils.ServiceInstanceKind:
+			m := &serviceModels.ServiceInstance{}
+			err := yaml.Unmarshal(doc, &m)
+			if err != nil {
+				return errors.Wrapf(err, "Error decoding policy document &s", string(doc))
```
neosab: Nit: s/policy/service instance

pkg/dispatchcli/cmd/delete_serviceinstance.go
```diff
@@ -0,0 +1,90 @@
...
+
+// NewCmdDeleteServiceInstance creates command responsible for deleting a service instance
+func NewCmdDeleteServiceInstance(out io.Writer, errOut io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "serviceinstance SERVICE_INSTANCE_NAME",
+		Short:   i18n.T("Delete service instance"),
+		Long:    deleteServiceInstanceLong,
+		Example: deleteServiceInstanceExample,
+		Args:    cobra.ExactArgs(1),
+		Aliases: []string{"apis"},
```
tenczar: Do you mean serviceinstances?

PR #358, opened 2018-04-06
Topic/ntenczar/cli update event

tenczar:


swagger/event-manager.yaml
```diff
@@ -647,6 +761,10 @@ definitions:
       name:
         type: string
         maxLength: 32
+      kind:
+        type: string
+        pattern: '^[\w\d\-]+$'
```
berndtj: really no reason that we have a pattern on a read-only field... no biggie though.

PR #360, opened 2018-04-06
Add Bearer Auth handler to other services

neosab:
Service accounts can login using bearer tokens which are validated by identity manager. However, we still need a dummy bearer token handler in all services else the request will get rejected by go-openapi runtime.

In the future, this handler func will validate token issued by identity-manager and will form a second layer of security.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/360)

### DIFF
--- a/pkg/api-manager/handlers.go
+++ b/pkg/api-manager/handlers.go
@@ -120,6 +120,12 @@ func (h *Handlers) ConfigureHandlers(routableAPI middleware.RoutableAPI) {
 		return token, nil
 	}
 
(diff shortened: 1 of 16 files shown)


examples/seed.yaml
```diff
@@ -45,7 +45,7 @@ tags:
   - key: role
     value: test
 ---
-kind: Function
+kind: unction
```
imikushin: Did you accidentally delete `F`?

neosab: Yup :D

PR #368, opened 2018-04-10
delete runs with function

berndtj:
* in the controller delete all runs before deleting the function
  - if this fails, it will be retried as the function hasn't been
    deleted
* remove the tags filtering on delete...not useful
* add e2e

### DIFF
--- a/pkg/function-manager/controller.go
+++ b/pkg/function-manager/controller.go
@@ -105,10 +105,21 @@ func (h *funcEntityHandler) Delete(obj entitystore.Entity) error {
 		return errors.Wrapf(err, "Driver error when deleting a FaaS function")
 	}
 
+	runs, err := getFilteredRuns(h.Store, &e.Name, nil)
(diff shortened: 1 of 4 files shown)


pkg/function-manager/controller.go
```diff
@@ -105,10 +105,21 @@ func (h *funcEntityHandler) Delete(obj entitystore.Entity) error {
 		return errors.Wrapf(err, "Driver error when deleting a FaaS function")
 	}
 
+	runs, err := getFilteredRuns(h.Store, &e.Name, nil)
```
neosab: If a run is in `INITIALIZED`/`CREATING` state, should wait for them to be `READY` before deleting them?

berndtj: No, but we should ensure that we get any orphaned functions in the controller sync and delete them appropriately from the FaaS.  We are not currently doing this.

PR #369, opened 2018-04-10
Make Dispatch work on Docker for Mac

imikushin:
To get around the local image registry availability problem, use
`host.docker.internal:5000` as the imageRegistry in dispatch install
config.yaml:

```yaml
  imageRegistry:
    name: host.docker.internal:5000
    insecure: true
```

Run a standalone image registry:
```
docker run -d -p 5000:5000 --restart=always --name=registry registry
```

In Docker for Mac preferences:
- add `host.docker.internal:5000` to the list of insecure registries
- add `http://localhost:5000` to the list of registry mirrors


kars7e:
Great! Can you also update the docs with your findings? E.g. add a "Docker on Mac" section

PR #377, opened 2018-04-13
Remove --enable-bootstrap-mode flag from Identity Manager

neosab:
Remove `--enable-bootstrap-mode` flag from identity manager. We will now watch for the bootstrap volume to be mounted, if mounted, we assume the bootstrap mode is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/377)


neosab:
@berndtj Unfortunately, the k8s client initialization failed in GKE environment. I fixed it, can you please re-approve?

Issue #359, opened 2018-04-06
Java language support


imikushin:
Base image for Dispatch providing support for the Java language. The base image sources are going to be in this repo: https://github.com/dispatchframework/java-base-image

- [ ] Bug
- [x] Feature
- [ ] Enhancement

## Detailed Description

To add Java language support, Dispatch users will have to create a base-image with Java support in their Dispatch installation. 

### Function programming model
A function in Java will be a class implementing `java.util.function.BiFunction<T,U,R>` interface. T, U and R are type parameters and should be represented by JSON-encodable types:

- T — context, best represented by Map<String,Object>
- U — payload, can be any JSON-encodable value type
- R — output payload, can be any JSON-encodable value type

Hello.java:
```java
    package com.example.functions;
    
    import java.util.Map;
    import java.util.function.BiFunction;
    
    public class Hello implements BiFunction<Map<String, Object>, Map<String, Object>, String> {
        public String apply(Map<String, Object> context, Map<String, Object> payload) {
            final Object name = payload.getOrDefault("name", "Someone");
            final Object place = payload.getOrDefault("place", "Somewhere");
            
            return String.format("Hello, %s from %s!", name, place);
        }
    }
```

## Context

Dispatch users will use the Java base-image to create images and functions:

```
$ dispatch create base-image java-base vmware/dispatch-java-base:0.0.1

$ dispatch create image my-java-libs java-base --runtime-deps ./hello-proj/pom.xml

$ dispatch create function my-java-libs hello1-java \
  ./hello-proj/src/java/com/example/functions/Hello.java
```

## Possible Implementation

The function-server in the base image can be implemented using e.g. [Undertow](https://github.com/undertow-io/undertow) as HTTP server. 

Function image Dockerfile should contain steps to build the Java class from the provided source file and install it into the local filesystem Maven repo in the function image.

## Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [x] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [x] High - Blocker

## Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

berndtj:
[...]

PR #385, opened 2018-04-17
Add java examples and e2e tests

rjew:
Examples and e2e tests for #359 

Once the image in dispatchframework/java-base-image/pull/1 is built and pushed to docker, e2e tests can be run to ensure that it is working correctly.

### DIFF
--- a/examples/java8/pom.xml
+++ b/examples/java8/pom.xml
@@ -1,0 +1,18 @@
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>io.dispatchframework.examples</groupId>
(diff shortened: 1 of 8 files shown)


examples/java8/pom.xml
```diff
@@ -0,0 +1,14 @@
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.github.vmware.dispatch.examples.java8.hellowithdeps</groupId>
```
tenczar: I might recommend cutting this down to just `github.vmware.dispatch.examples`

rjew: Sure I'll make it shorter

PR #388, opened 2018-04-19
Update auth docs

zimengyang:
Update authentication docs


zimengyang:
Added note of using github user's email as subject while creating policy (not user name)

PR #405, opened 2018-05-02
Add boolean config support and secret as env variable for event manager

zimengyang:
* Add boolean config support for event manager
* Use secrets as environment variable for containers


pkg/dispatchcli/cmd/create_eventdriver.go
```diff
@@ -64,13 +64,18 @@ func createEventDriver(out, errOut io.Writer, cmd *cobra.Command, args []string)
 	var driverConfig []*models.Config
 	for _, conf := range createEventDriverConfig {
 		result := strings.Split(conf, "=")
-		if len(result) != 2 {
-			fmt.Fprint(errOut, "Invalid Configuration Format, should be --config key=value")
+		if len(result) == 1 {
```
kars7e: you can do this as:
```go 
switch len(result):
case 1:
case 2:
default
```

it should be a bit more readable

pkg/event-manager/drivers/k8s_backend.go
```diff
@@ -92,9 +92,11 @@ func (k *k8sBackend) makeDeploymentSpec(driver *entities.Driver) (*v1beta1.Deplo
 		inputMap[key] = val
 	}
 
-	// secrets are more important than config
+	// holds secrets for event driver (as env variable in container)
+	secretMap := map[string]string{}
```
kars7e: `secrets` is already a `map[string]string`, so seems like you don't need that extra map

zimengyang: Thanks! Just update it

PR #410, opened 2018-05-04
Initial support for Jaeger & OpenTracing

kars7e:
Jaeger can be installed optionally. To install, add
```
jaeger:
  enabled: true
```
to install.yaml

**Changes**:
* adds `jaeger-client` dependency
* adds optional `--tracer` flag to every service
* adds utility function/wrapper to create a tracer
* adds jaeger installation logic to `dispatch install`
* adds HTTP middleware to inject tracing context
* sets minimal supported version of Kafka to 0.11 to support message headers

**Testing**:
Manually tested that jaeger works (once db situation is resolved, we can enable jaeger in every deployment)


pkg/middleware/tracing.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"net/http"
+
+	"github.com/justinas/alice"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+)
+
+// NO TESTS
```
berndtj: We should start adding comments as to why we aren't adding tests for a particular module.  I'm not really disagreeing with you here, but we don't want to appear sloppy.

PR #411, opened 2018-05-04
[e2e/java] Update function tests to test Spring support

tenczar:
Creates a new java image with Spring dependencies. Executes a Spring annotated function.


examples/java8/HelloSpring.java
```diff
@@ -0,0 +1,73 @@
+/**
+ *
```
berndtj: If we're going to have comments, they should be filled out (look at hello.py for reference).  Otherwise it looks sloppy.  We also should add comments to the *.js examples, but that's a separate issue.

PR #427, opened 2018-05-10
Letsencrypt support via cert-manager

berndtj:
This is a work in progress to support real certificates via
letsencrypt.  Here are the notes:

* Requires DNS challenge (via route53 for now) as kong is not an
  ingress controller and isn't compatible with the HTTP challenge
  scheme
    - HTTP would be preferred as we don't need to depend on access
      to a 3rd party DNS service (route53)

* I'm not sure we should package this up together.  This support
  may be better handled via manually installing additinal charts
  (cert-manager, certificate) as the failure modes are hard to
  debug and handle automatically

* User must manually set a secret with AWS IAM secret:

```kubectl create secret generic route53 --namespace kube-system --from-literal secret-access-key=$AWS_SECRET_ACCESS_KEY```

* Please comment!

Here's a working install.yaml:

```
ingress:
  serviceType: LoadBalancer
apiGateway:
  serviceType: LoadBalancer
  host: apibjung03.dispatchframework.io
letsEncrypt:
  email: bjung@vmware.com
  staging: false
  route53:
    accessKeyID: ***********
dispatch:
  host: bjung03.dispatchframework.io
  port: 443
  tls:
    ca: letsEncrypt
  faas: riff
  eventTransport: kafka
  bootstrapUser: bjung@vmware.com
  oauth2Proxy:
    provider: github
    clientID: **********
    clientSecret: ***********
```

### CODE REVIEW

File: docs/_guides/gke.md [hunk 1]
```diff
@@ -0,0 +1,174 @@
...
+
+```
+kubectl create clusterrolebinding tiller-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:default
+```
+
+The following command it to work around a known [helm](https://github.com/kubernetes/helm/issues/3409)
+[issues](https://github.com/kubernetes/helm/issues/3379).  If it fails, retry:
+
+```
+helm init --tiller-image powerhome/tiller:git-3b22ecd --wait
```
kars7e: This shouldn't be needed with helm 2.8.2, as both of those issues were addressed (and it shouldn't require retry either). I will update our CI to use 2.8.2.

berndtj: good to know.

PR #428, opened 2018-05-11
[e2e/java] Update java image name in tests

tenczar:
The java base image has been updated to use java 10. The image name has dropped the java8 label to show that this can be used with any current java application.


tenczar:
This has been resolved by #470

Issue #431, opened 2018-05-11
Dispatch does not work with artifactory image registry


berndtj:
## Bug Report
When the image registry is configured to point to an artifactory image registry, functions cannot be created.

```
time="2018-05-11T10:51:45Z" level=error msg="Driver error when creating a FaaS function: Error building image for function 'e8aa7160-11b2-449a-ab81-5a93a3d5ab9b': failed to pull image 'symphony-docker-local.jfrog.io/docker-local/dispatch-test/b31e30ba-2f9e-43cd-882a-09e086c48c25:latest': Error response from daemon: Get https://symphony-docker-local.jfrog.io/v2/docker-local/dispatch-test/b31e30ba-2f9e-43cd-882a-09e086c48c25/manifests/latest: unknown: Authentication is required"
```

### Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [x] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

### Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

rbtcollins:
This looks like a trivial lack of registry secret provisioning. 

To prepare the namespace:
https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-in-the-cluster-that-holds-your-authorization-token

and then for each pod (deployment/job/statefulset):
https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret

berndtj:
Thanks, it's not quite that simple :).  We need to propagate this through our install.  Also, the FaaS needs to support image pull secrets (openfaas does).  Lastly, because of changes in this format between k8s 1.8 and 1.9, things don't always "just work".

berndtj:
Should have a patch very soon though

Issue #435, opened 2018-05-14
Dispatch does not support private registries


berndtj:
## Bug Report
While dispatch works to push images to registries which require authentication, it does not support pulling images from private repositories.

### Expected behavior

When configuring a repository, dispatch should propagate the necessary secrets to any entity which requires image pull access.

### Current behavior

Functions cannot be deployed.

### Steps to reproduce

Use a private repo.

### Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [x] High - Blocker

### Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

kars7e:
Isn't that fixed by #438 ?

berndtj:
Yes, this is fixed.

Issue #430, opened 2018-05-11
Transport does not obey "single namespace" option


berndtj:
## Bug Report
Deployed dispatch with single namespace option:
```
dispatch install -f cascade-install.yaml --single-namespace default
```

Transport still installed in "dispatch" namespace:
```
helm list
NAME       	REVISION	UPDATED                 	STATUS  	CHART               	NAMESPACE
api-gateway	1       	Fri May 11 10:19:40 2018	DEPLOYED	kong-0.1.2          	default  
dispatch   	1       	Fri May 11 10:21:40 2018	DEPLOYED	dispatch-0.1.13     	default  
ingress    	2       	Fri May 11 10:19:32 2018	DEPLOYED	nginx-ingress-0.19.0	default  
openfaas   	1       	Fri May 11 10:21:22 2018	DEPLOYED	openfaas-1.0.0      	default  
postgres   	1       	Fri May 11 10:19:37 2018	DEPLOYED	postgresql-0.8.5    	default  
transport  	1       	Fri May 11 10:21:36 2018	DEPLOYED	kafka-0.0.1         	dispatch
```

### Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [x] High - Blocker

### Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

kars7e:
Fixed by #438

PR #438, opened 2018-05-14
Adding support for private registries (image pull)

berndtj:
* Previously functions could not pull private images
* Also fix single-namespace install
  - transports now install correctly
* Only supports OpenFaaS
  - Riff does not support image pull secrets

Needs e2e tests.  Tested manually on minikube

Fix: #431 #430


berndtj:
fix #435

PR #439, opened 2018-05-15
Event manager deploy and sync improvement

zimengyang:
* Add backoff logic for Deploying eventdriver, status will be set to READY only when deployment has available replica
* Sync with underlying k8s deployment and pod to check availabilities for eventdriver and report error state
* Add eventdriver error types and error reason
* Partially add error handling
* Add e2e test for eventdriver creation error

* **TODO**: 
  * Better error handling 
  * Find better way to get k8s Pod status. 

```bash
dispatch get eventdriver
         NAME         |     TYPE     | STATUS | SECRETS | CONFIG |                     REASON
--------------------------------------------------------------------------------------------------------------------
  open-javelin-636794 | aws-no-image | ERROR  |         |        | DeploymentNotAvailable, k8s: deployment
                      |              |        |         |        | `event-driver-aws-no-image-open-javelin-636794`
                      |              |        |         |        | is `waiting`: ImagePullBackOff, Back-off
                      |              |        |         |        | pulling image "dispatch-events-no-image"
--------------------------------------------------------------------------------------------------------------------
```


pkg/event-manager/drivers/k8s_backend.go
```diff
@@ -237,10 +271,40 @@ func (k *k8sBackend) Delete(ctx context.Context, driver *entities.Driver) error
...
+				}
+			}
 
+		}
+	}
+
+	return nil
+}
+
+func (k *k8sBackend) Update(ctx context.Context, driver *entities.Driver) error {
```
berndtj: Make sure to add a comment on all public methods (we need to keep running lint)

zimengyang: Yeah, sure, i will make sure and double check it. For this case, golint wasn't complaining since `k8sBackend` was treated as unexported, it's methods were treated as unexported too.

Issue #429, opened 2018-05-11
Add Kubeless as FaaS driver


andresmgot:
### Detailed Description

As far as I understand, possible function drivers are OpenFaas, Openwhisk and riff. We would like to contribute with a Kubeless driver.

### Context

[Kubeless](https://github.com/kubeless/kubeless) is one important FaaS platform for Kubernetes so it would be a good fit for dispatch.

### Possible Implementation

There is not much documentation on how to implement a new FaaS driver so it would be great to have some help here.

### Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [X] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

kars7e:
Hey @andresmgot, that would be awesome. Indeed we don't have any docs for adding new driver - I'd love to have this as an exercise to add some! Let me list the things that might be useful for you, and we can go from there.
* Currently, each functions driver lives in a separate directory within [`functions`](https://github.com/vmware/dispatch/tree/master/pkg/functions) folder (we had discussions about pulling them out into separate directory structure/repositories, but we don't have any concrete plans yet, so we can add kubeless there too).
* Each driver must implement [`FaaSDriver`](https://github.com/vmware/dispatch/blob/186129d49e735fbd5994ba995faded7bc343497b/pkg/functions/types.go#L61) interface. 
* There is a driver map which configures each driver in [`main.go` file](https://github.com/vmware/dispatch/blob/186129d49e735fbd5994ba995faded7bc343497b/cmd/function-manager/main.go#L38). This will be refactored, but don't worry about that.
* You can look at one of the existing drivers, e.g. [OpenFaaS](https://github.com/vmware/dispatch/blob/186129d49e735fbd5994ba995faded7bc343497b/pkg/functions/openfaas/driver.go), to get a general idea behind what driver usually does.
[...]

PR #440, opened 2018-05-15
[WIP] Kubeless driver

andresmgot:
Ref #429 

This is the first draft for the Kubeless driver. Right now there are two commits in this PR:
 - 6cede5c Update vendor to include kubeless (done through `dep ensure`, let me know if there is a better way to do it)
 - 392680b Actual changes to introduce Kubeless

These some of the missing items to implement:
 - [X] Kubeless installation
 - [X] Schema validation
 - [X] Unit tests
 - [ ] e2e tests


pkg/function-manager/handlers.go
```diff
@@ -121,9 +122,14 @@ func functionModelOntoEntity(m *v1.Function, e *functions.Function) error {
 	if m.Main != nil && *m.Main != "" {
 		main = *m.Main
 	}
+	if m.Image != nil && *m.Image != "" {
+		e.ImageName = *m.Image
+	}
+	if m.Runtime != nil && *m.Runtime != "" {
```
tenczar: We will likely want validation somewhere that fails if Image and Runtime are empty.

imikushin:
Hi @andresmgot, can you squash your vendoring commits into one?

pkg/function-manager/handlers.go
andresmgot: added in `controller.go` in the `Add` method

andresmgot:
> My biggest concern is that we use the images to provide end users a means to add language support without us adding it directly. Ivan will likely have more thoughts on how we can address this.

AFAIK that is still possible, I just made the parameter `image` optional. Am I wrong?

> Hi @andresmgot, can you squash your vendoring commits into one?

Done (after some trouble when rebasing master)

imikushin:
Thanks for squashing the vendor, @andresmgot! It really helps reviewing.

Can we do without adding `Runtime` field? 

The reason here is: in Dispatch, to add a language runtime you only need to add the [base image for the language](https://vmware.github.io/dispatch/documentation/specs/language-support). For a how to guide, check out this post in the Dispatch blog: [Adding a Language to Dispatch](https://medium.com/dispatchframework/adding-a-language-to-dispatch-3515cce08e9f). 

I have checked [Kubeless Runtimes docs](https://kubeless.io/docs/runtimes/) and found this should be possible.

andresmgot:
oh, I got it wrong, I didn't fully understand the difference between `base-image` and `image`. So after doing some analogy:

 - Your `base-image` is the same concept than our `runtime`. The generic image that contains the runtime binaries. With the difference that our `base-image` already contains what you call the `function-server`. See the [nodejs runtime](https://github.com/kubeless/kubeless/blob/master/docker/runtime/nodejs/Dockerfile.6).
 - What you do in your `image` (installing external dependencies of the functions) we do it in an `initContainer` ([ref](https://github.com/kubeless/kubeless/blob/master/pkg/langruntime/langruntime.go#L188))
 - What you do in the `function-template` (copying the function content) we also do it in a different init container ([ref](https://github.com/kubeless/kubeless/blob/master/pkg/utils/kubelessutil.go#L53)).
 - Finally you store the result as a different image in a Docker registry. We also do that but only optionally if the user enables that feature.

Since we don't have the concept of `image` Kubeless need the `runtime` to know what will be the build process of a specific function. The other option will be to understand `image` as `base-image` in the case of Kubeless but I don't know if that will be too confusing. I mean something like:

```yaml
kind: BaseImage
name: nodejs6-base
dockerUrl: kubeless/nodejs:0.0.3
language: nodejs6
---
kind: Function
name: hello-js
code: '@nodejs6/hello.js'
image: nodejs6-base
```

What do you think about that?

berndtj:
@andresmgot I think you've got it for the most part.  Our base image should also contain the "function server" (like yours).  In fact we don't have FaaS specific base images, each base image contains everything it needs for *all* supported FaaS.  @imikushin can explain better.  Therefore, we won't have a specific kubeless base image in the end.

imikushin:
Our goal was to decouple language runtimes from FaaS engines. 

Base images, images and functions are all Docker images:
- Base-image's responsibility is to provide the container OS, the language runtime and templates for building _images_  and _functions_. 
- Images allow us to manage system (OS shared libraries and binaries) and runtime dependencies (language libraries). 
- Functions get instantiated on demand and serve invocation requests.

A _function image_  (the result of calling `d.imageBuilder.BuildImage(f)`) is a Docker image that can be used to spin up function instances. It should only serve the Function Runtime API on the specified `PORT` (by default, 8080). It's up to the FaaS driver: how to spin up function containers (instances of the _function image_) and how to bring function invocation requests to an instance. 

Dispatch functions can be treated by Kubeless as functions with custom images. For example, if a function `f` is being created with `f.FaasID == "f00ba1001"` and it's _function-image_ DockerURL is `host/hello1-cafe1001`, then the Kubeless FaaS driver would perform an equivalent of this CLI command:
```bash
kubeless function deploy --runtime-image host/fn-cafe1001 f00ba1001
```
I hope this helps! Don't hesitate to DM me for any help.

andresmgot:
True, that's a valid option as well. Thanks for clarifying! I will try that.

kars7e:
This is looking really good! I'll let @imikushin review it, I have only one ask, @andresmgot can you re-run `dep ensure` without running `prune`? #441 added prune inside `Gopkg.toml` and is executed when you run `dep ensure`. Dep version is maintaining license-related files, let's keep them

andresmgot:
done @kars7e I also updated the code to fix some conflicts with master

berndtj:
Nice job @andresmgot!

pkg/functions/kubeless/driver.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (err *systemError) StackTrace() errors.StackTrace {
+	if e, ok := err.Err.(functions.StackTracer); ok {
+		return e.StackTrace()
+	}
+
+	return nil
+}
+
+// New creates a new OpenFaaS driver
```
berndtj: s/OpenFaaS/Kubeless/

pkg/functions/kubeless/driver.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return list.Items[0].Name, nil
+}
+
+func (d *kubelessDriver) getHTTPReq(funcName, eventID, eventNamespace string, body []byte) (*http.Request, error) {
+	svc, err := d.services.Get(funcName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Unable to find the service for function %s", funcName)
+	}
+	funcPort := strconv.Itoa(int(svc.Spec.Ports[0].Port))
```
imikushin: We know the port in advance: it's `8080`. I'd make it a constant.

andresmgot:
thank you for the reviews @imikushin @berndtj! It's more simple now.

berndtj:
Thanks @andresmgot Dispatch now supports kubeless!

PR #441, opened 2018-05-15
Add prune to Gopkg.toml

kars7e:
We've been using separate prune command, but that's not obvious for external contributors. This is an attempt to use "prune" built in the dep itself. It deletes even more files than external prune command, let's tune if needed.


neosab:
Cool

Issue #416, opened 2018-05-07
Using directories to build functions


imikushin:
This was requested by many potential users. Most of these requests were in the form of questions: “I need to create a function that needs a few additional files. How do I go about it? What if some of my dependency packages are not available from an artifact repository?”

## Creating functions after the change

When creating a function, the users:

1. Pass a directory (perhaps containing the function source code, resources and supporting code) — called the *source directory*.
2. Specify a language-specific *function module*, e.g.:
  1. a fully qualified class name — for Java
  2. a module name (an argument to `require`) — for JavaScript
  3. a package name — for Python

Dispatch puts the directory contents into the *function image*, builds (if necessary for the language) the specified *function module* and sets it as the function to run.

### `dispatch create function`

The mandatory `FUNCTION_FILE` argument is replaced with the optional `DIR` argument and the mandatory `--module` parameter:

****
    dispatch create function $IMAGE_NAME $FUNCTION_NAME 
      [$DIR] --module=$lang_specific_string \
      [--schema-in SCHEMA_FILE] \
      [--schema-out SCHEMA_FILE] \
      [flags]

## API changes

There is a change in the parts of Function Manager API:

- `POST /function` request body
- `GET /function/{functionName}` response body

Fields removed:

- `code` — the function code is provided as part of the source directory
- `main` — no longer used

Fields added:

- `dir` — gzipped tar-ball of the function’s *source directory*
- `module` — language specific *function module* name (fully qualified name or path)

markpeek:
Thanks for writing this up. A couple of comments:
- should we allow zip archives (ala AWS lambda)?
- should allow for multiple directories or zip files to be specified?
- should we allow specifying a base directory to place or unpack each artifact?

berndtj:
I'm not sure in what cases "dir" would be an optional argument.  Also, can it just be "path" which may be a file or directory.

I see that AWS does indeed use a handler arg the same way as specified here.  I guess I don't have a huge objection to requiring it, though maybe we could default to the "handler" functions that support today.

imikushin:
If we change "dir" for "path" (as in, not only dirs, but also files), the `--module` will be optional: there's no need to specify it if there's just one source file.

Zip files: sure, why not!

Multiple archives / base dirs: why? would sub-dirs work? it's just a single function after all.

[...]

PR #469, opened 2018-05-21
[cli/function] Update create function cli param order for consistency

tenczar:
Previously the image name was the first required parameter to dispatch create function as in
`dispatch create function my-image my-function function.py`

The params are now ordered like
`dispatch create function my-function my-image function.py`


imikushin:
This is addressed in #470, according to the spec in #416.

tenczar:
Closing based on previous comment. Fixed in #470

PR #470, opened 2018-05-22
Use sourceDir and handler to build functions

imikushin:
Implement #416 

Get rid of Exec when creating functions and building function images.
Build function images *before* calling faasDriver.Create().
Write the source dir.
rm function.Code field.
Move TarDir and Untar to tar.go
Add TarGzBytes() to utils/tar.go
CLI: create function.
Function.sourcePath for seed.yaml
TarDir to correctly normalize paths.
Migrate e2e tests.


examples/python3/hello.py
```diff
@@ -13,7 +13,7 @@
 dispatch create image python3 python3
 
 Create a function:
-dispatch create function python3 hello-python examples/python3/hello.py
+dispatch create function --image=python3 hello-python examples/python3 --handler=hello.handle
```
berndtj: I really, really want to also preserve the existing support for simply specifying a filename as a function (without handler).  If I specify a single file (i.e. examples/python3/hello.py) it should be implicit that we are going to use --handler hello.handle.  We don't even need to do anything fancy to support this, it could all be hidden in the CLI.

berndtj: OK, I still haven't had my coffee :).  But the CLI could take care of simply tar'ing up a single file in the "current" usage pattern, but the base image would likely have to call the "default handler" as it's very language specific.  Anyway, I'd like to gauge the effort required to support this.

PR #473, opened 2018-05-23
Remove static OrgID across services

neosab:
Until now, all Dispatch services have a static OrgID flag that sets the orgID on entities created by those services. With this change, orgID is per-request scoped and passed as a custom request header *X-DISPATCH-ORG-ID* to all API's. As a result, the services can handle multiple tenants in the future. Identity Manager is responsible to set this header for all user requests while individual services set this header for inter-service requests. This patch is a pre-requiste to introduce multi-tenancy in Dispatch.

* Removes OrgID flags from most services
   *Note*:- We don't remove it from Identity Manager since we don't have tenancy concept. Hence, it's statically set only in the identity manager. We also don't remove from service-manager since the k8s service catalog lacks tenancy constructs.

* Set the *X-DISPATCH-ORG-ID* as required req header for API's

* Set the *X-DISPATCH-ORG-ID* in auth response header from IDM

* Introduce a temporary entity store method ListOrgID's to retrieve all org's. This will be replaced by CRUD methods that handle org lifecycle. Currently, this method is hardcoded to return a single tenant.

* Fix controller sync to handle entities across all org's. 

* Enhance service clients to handle the org header

* Fix Charts

* Fix tests

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/473)


pkg/application-manager/handlers.go
```diff
@@ -223,6 +221,7 @@ func (h *Handlers) updateApp(params application.UpdateAppParams, principal inter
 	}
 	e.Status = entitystore.StatusREADY
 	updatedEntity := applicationModelOntoEntity(params.Body)
+	updatedEntity.OrganizationID = params.XDISPATCHORGID
```
tenczar: Do we really want to update the organization id? I imagine that for security reasons we wouldn't want one object to become a part of a different organization. Users should recreate the resource in the new org.

pkg/client/images.go
```diff
@@ -22,22 +22,22 @@ import (
 // ImagesClient defines the image client interface
 type ImagesClient interface {
 	// Images
-	CreateImagectx(ctx context.Context, image *v1.Image) (*v1.Image, error)
```
tenczar: This seems like this was a bug. Did we really never try to use the ImagesClient interface? The 'implementing' struct seems to have always used `CreateImage` instead of `CreateImagesctx`

pkg/image-manager/handlers.go
```diff
@@ -219,6 +216,7 @@ func (h *Handlers) addBaseImage(params baseimage.AddBaseImageParams, principal i
 
 	baseImageRequest := params.Body
 	e := baseImageModelToEntity(baseImageRequest)
+	e.OrganizationID = params.XDISPATCHORGID
```
tenczar: same question about being able to change this.

pkg/client/images.go
kars7e: Yeah that's a bug, we haven't really used this interface yet, as function manager requires only ’GetImage’, and declares it's own interface for that. This interface would eventually be used in CLI

pkg/application-manager/handlers.go
neosab: It looks misleading but we don't allow an entity to be updated with a different orgID. We query if the entity exists within the org specified by the user and then update it. I could do `updatedEntity.OrganizationID = e.OrganizationID` to be less ambiguous.

pkg/client/images.go
neosab: Yeah, we only use the GetImage method in Function Manager.

pkg/application-manager/handlers.go
berndtj: Perhaps even less ambiguous would be if we added the orgID to the model, so that we don't need to tack it on at all.

pkg/client/images.go
```diff
@@ -22,22 +22,22 @@ import (
...
-	DeleteImage(ctx context.Context, imageName string) (*v1.Image, error)
-	UpdateImage(ctx context.Context, image *v1.Image) (*v1.Image, error)
-	GetImage(ctx context.Context, imageName string) (*v1.Image, error)
-	ListImages(ctx context.Context) ([]v1.Image, error)
+	CreateImage(ctx context.Context, organizationID string, image *v1.Image) (*v1.Image, error)
```
berndtj: Does org ID belong in the context rather than an arg?

neosab:
> Wow, a lot of work went in here! Very much needed change. Just some suggestions:
> 
> all caps header seems weird, couldn't it just be X-Dispatch-Org? You could probably drop the "-Id" as well.
> I'm not sure about having it as a header instead of part of the path/request, but I can't find any good reason for it so let it be :-)

Wov, this is exactly what I was contemplating for quite a while. 

TLDR; I would like the tenant id in the URI when we add the multi-tenancy support. It's kinda easy to switch to from header to path. 

I actually wanted a new JWT token (subsuming the oauth2proxy cookie) that is issued by the IAM describing the tenant info to be passed in a header (à la `X-AUTH-TOKEN`). And hence I felt it was redundant to include the tenant-id again in the path. But it quickly turned out to be tedious since we also make authorization decisions centrally in the IAM and we need to define scopes in the token to describe privileges. I then pivoted to passing ORG-ID's in the header for the time being. Once we have multi-tenancy support, having tenant-id in the URI will be more meaninful.

> Looking at pkg/client changes, Is specifying organizationID for every request really needed? Could it be specified client-wide when client is created, and then optionally provided? (e.g. via Functional options). It's probably always needed when our services talk internally, I'm trying to think from the possible SDK angle. Also there might be more request-specific options that user might set in future, we can't address them all now but I wonder if we could reduce future changes.

Yeah didn't think too much from the SDK perspective. We have some options :- 
1. Make orgID optional since it can be set during client initialization.
2. Pass orgID in ctx but I don't want to abuse context for optional arguments to the function. 
3. Add an options field (again a dignified way instead of passing args through context)
4. Functional Options, hmm, let me take a look

neosab:
I haven't yet fixed the client interface part. I will do that in a subsequent patch. Other than that, I think we can merge this one.

pkg/client/images.go
neosab: I will make it easy to pass orgID in the clients in a subsequent PR. 

On the other comment, I could add orgID to model but since orgID is passed as a header in the request, we still have to set it explicitly on the model (since the request body won't include the orgID).

PR #476, opened 2018-05-24
Retry stalled/failed image building

berndtj:
* Retry image creation if entity is still INITIALIZED after timeout
  - Also will pick up in progress or interrupted work across restarts
* Fix installer to work with both IPs and ELB/DNS names for exposed services
* Add some additonal logging to image manager
* Set resource requirements for image builder container
* Pin DIND to version

### DIFF
--- a/charts/dispatch/charts/image-manager/values.yaml
+++ b/charts/dispatch/charts/image-manager/values.yaml
@@ -26,20 +26,21 @@ ingress:
   tls: {}
     # Secrets must be manually created in the namespace.
     # secretName: dispatch-tls
-resources: {}
+resources:
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
   # resources, such as Minikube. If you do want to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-  # limits:
-  #  cpu: 100m
-  #  memory: 128Mi
-  #requests:
-  #  cpu: 100m
-  #  memory: 128Mi
+  limits:
+    cpu: 500m
+    memory: 4Gi
+  requests:
+    cpu: 250m
+    memory: 2Gi
 registry: {}
   # insecure: false
   # uri: docker-docker-registry.docker.svc.cluster.local:5000
 data:
   # persist: false
   hostPath: /var/image-manager
+# debug: false
--- a/pkg/dispatchcli/cmd/log.go
+++ b/pkg/dispatchcli/cmd/log.go
@@ -57,6 +57,7 @@ func NewCmdLog(out, errOut io.Writer) *cobra.Command {
 		Long:    logLong,
 		Example: logExample,
 		Args:    cobra.MinimumNArgs(1),
+		Aliases: []string{"logs"},
 		Run: func(cmd *cobra.Command, args []string) {
 			err := runLog(out, errOut, cmd, args)
 			CheckErr(err)
--- a/pkg/image-manager/controller.go
+++ b/pkg/image-manager/controller.go
@@ -8,6 +8,7 @@ package imagemanager
 import (
 	"context"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -21,6 +22,7 @@ import (
 // ControllerConfig defines the image manager controller configuration
 type ControllerConfig struct {
 	ResyncPeriod   time.Duration
+	Timeout        time.Duration
 	OrganizationID string
 }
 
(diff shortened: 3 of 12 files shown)


charts/dispatch/charts/image-manager/values.yaml
```diff
@@ -26,20 +26,21 @@ ingress:
...
   # choice for the user. This also increases chances charts run on environments with little
   # resources, such as Minikube. If you do want to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-  # limits:
-  #  cpu: 100m
-  #  memory: 128Mi
-  #requests:
-  #  cpu: 100m
-  #  memory: 128Mi
+  limits:
```
tenczar: We might want to change the above comment block if we are going to be uncommenting these fields.

pkg/dispatchcli/cmd/log.go
```diff
@@ -57,6 +57,7 @@ func NewCmdLog(out, errOut io.Writer) *cobra.Command {
 		Long:    logLong,
 		Example: logExample,
 		Args:    cobra.MinimumNArgs(1),
+		Aliases: []string{"logs"},
```
tenczar: thanks

pkg/image-manager/controller.go
```diff
@@ -94,6 +96,42 @@ func (h *baseImageEntityHandler) Error(ctx context.Context, obj entitystore.Enti
...
+	timeout time.Duration
+}
+
+// CheckAndLock checks to see if the entity is currently being worked on if not
+// stores the entity
+func (h *imageEntityHandler) CheckAndLock(obj entitystore.Entity) bool {
+	// Only do a single image create at a time (for a given image)
+	now := time.Now()
+	v, loaded := h.lock.LoadOrStore(obj.GetID(), now)
+	if loaded {
```
kars7e: NIT: I know it's a matter of preference, but handling `if !loaded` first would reduce the overall indentation. Same with nested if, returning false in `if duration > h.timeout` would remove the `else` clause and increase readability.

pkg/image-manager/controller.go
```diff
@@ -113,12 +151,22 @@ func (h *imageEntityHandler) Add(ctx context.Context, obj entitystore.Entity) (e
...
+	if h.CheckAndLock(i) {
+		return nil
+	}
+
+	defer h.Unlock(i)
 	defer func() { h.Store.UpdateWithError(ctx, i, err) }()
 
+	log.Infof("Creating image %s/%s", i.OrganizationID, i.Name)
 	if err = h.Builder.imageCreate(ctx, i, &bi); err != nil {
+		i.Status = entitystore.StatusERROR
```
kars7e: `UpdateWithError` already checks if err is not nil and sets the status accordingly, so this seems redundant.

pkg/image-manager/controller.go
```diff
@@ -135,6 +183,9 @@ func (h *imageEntityHandler) Delete(ctx context.Context, obj entitystore.Entity)
 
 	i := obj.(*Image)
 
+	h.ForceLock(i)
```
kars7e: This seems like a race condition, if there is an `Add` operation in progress and we deleted the image in between, `h.Store.UpdateWithError` will try to update a non-existent entity, is that safe?

pkg/image-manager/image_builder.go
```diff
@@ -312,25 +317,30 @@ func (b *ImageBuilder) imageCreate(ctx context.Context, image *Image, baseImage
...
 	}
 
 	if err := b.copyImageTemplate(tmpDir, baseImage.DockerURL); err != nil {
+		log.Errorf("failed to copy image template from image %s: %s", image.Name, err)
 		return err
 	}
 
 	spFile := filepath.Join(tmpDir, systemPackagesFile)
 	if err := b.writeSystemPackagesFile(spFile, image); err != nil {
+		log.Errorf("failed to write packages file for image %s [%s]: %s", image.Name, spFile, err)
```
kars7e: I feel like we need a `LogAndWrap(err error, message string, format... interface{}) error` function :-)

pkg/images/builder.go
```diff
@@ -76,8 +76,8 @@ func Build(ctx context.Context, client docker.ImageAPIClient, dir, name string,
 		return errors.Wrap(err, "failed to create a tarball archive")
 	}
 
-	log.Debugf("Building image %s from tarball", name)
-	r, err := client.ImageBuild(ctx, tarBall, types.ImageBuildOptions{
+	log.Infof("Building image %s from tarball", name)
+	r, err := client.ImageBuild(context.Background(), tarBall, types.ImageBuildOptions{
```
kars7e: Was this intentional? why is this needed?

berndtj: not intentional at all

berndtj:
I'm going to redo this patch.  While it is solving an issue, it's rushed.  It didn't solve the actual problem I am dealing with.

PR #478, opened 2018-05-25
Add switchable contexts for config

berndtj:
* keep map of configurations with one current context
* list contexts (highlight current)
* set current
* push the certificate chart

### DIFF
--- a/pkg/dispatchcli/cmd/manage.go
+++ b/pkg/dispatchcli/cmd/manage.go
@@ -34,13 +34,13 @@ var (
 
 	manageExample = i18n.T(`
 # Enable Dispatch bootstrap mode while specifying bootstrap user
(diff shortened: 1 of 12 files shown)


pkg/dispatchcli/cmd/manage.go
```diff
@@ -34,13 +34,13 @@ var (
 
 	manageExample = i18n.T(`
 # Enable Dispatch bootstrap mode while specifying bootstrap user
-dispatch manage --enable-bootstrap-mode --bootstrap-user admin@example.com -f ./config.yaml
```
neosab: I think we should move the enable/disable bootstrap flags to their own subcommands in another PR.

berndtj:
I agree @markpeek, this was a bit quick and dirty and also not meant to be the "full" solution.  We can iterate on it.

PR #480, opened 2018-05-25
 Misc install and chart fixes

berndtj:
* set default resource requirements for dind container
* get the hostname (ELB) address back for installation on AWS
* allow image manager specific debug flag (quiet docker output)
* logs alias
* vcenter driver logs (temporary implementation)


pkg/dispatchcli/cmd/create.go
```diff
@@ -105,6 +119,11 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 			if err != nil {
 				return errors.Wrapf(err, "Error decoding image document %s", string(doc))
 			}
+			manifest, err := resolveFileReference(m.RuntimeDependencies.Manifest)
```
tenczar: Is there an example of this somewhere?

berndtj: It's in my own demo, but I should probably add something to seed, so continues to work.

berndtj:
This looks to have broken something.  Needs e2e pass before merging

PR #498, opened 2018-06-06
super minor changes to docs.

nickaashoek:
Changed minikube documentation to reflect that some of those commands were incorporated into the default settings.

Updated gke docs to fix the helm init command (--tiller-image didn't specify an image).


docs/_guides/minikube.md
```diff
@@ -42,7 +42,7 @@ will not have access to the private network you are connecting to with the VPN.
 3. Create a kubernetes cluster:
 
 ```bash
-$ minikube start --vm-driver=hyperkit --bootstrapper=kubeadm --disk-size=50g --memory=6144
+$ minikube start --vm-driver=hyperkit
```
berndtj: I think it's a good idea to keep the disk size and memory requirements.  We've had issues before.

PR #502, opened 2018-06-07
Use client SDK for secrets and services in CLI

berndtj:
* Also start adding unittests for the above CLI functionality
* Get the error cause for API errors


berndtj:
Looks like update is broken now... will look into it

PR #506, opened 2018-06-12
Remove OrgID config flags in some services

neosab:
Few services were missed in the previous refactor that removed the static OrgID's from service configuration. This patch also fixes other unit tests that weren't passing OrgID to the store methods.

### DIFF
--- a/pkg/event-manager/drivers/http_handlers.go
+++ b/pkg/event-manager/drivers/http_handlers.go
@@ -51,7 +51,6 @@ type ConfigOpts struct {
 	Tracer          string
 	K8sConfig       string
 	DriverNamespace string
-	OrgID           string
 }
 
 // NewHandlers Creates new instance of driver handlers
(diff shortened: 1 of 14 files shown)


pkg/event-manager/drivers/http_handlers.go
```diff
@@ -297,14 +296,14 @@ func (h *Handlers) deleteDriver(params driverapi.DeleteDriverParams, principal i
 	return driverapi.NewDeleteDriverOK().WithPayload(d.ToModel())
 }
 
-func (h *Handlers) getDT(ctx context.Context, driverTypeName string) *entities.DriverType {
+func (h *Handlers) getDT(ctx context.Context, ordID string, driverTypeName string) *entities.DriverType {
```
kars7e: typo? ordID/orgID

neosab: of course!

PR #507, opened 2018-06-12
CLI: dispatch create seed-images

imikushin:
When used without any arguments, creates base-images and images from
images.yaml at the same revision as the CLI build. When used with
`-O file.yaml` flag, writes the YAML to the specified file (nothing gets
created).

Also:
- Bump base-image versions
- Move images.yaml to the repo root


Makefile
```diff
@@ -14,12 +14,13 @@ SWAGGER := $(GOPATH)/bin/swagger
...
+CLICMD_PACKAGE := github.com/vmware/dispatch/pkg/dispatchcli/cmd
 
-GO_LDFLAGS :="
-GO_LDFLAGS += -X $(VERSION_PACKAGE).version=$(VERSION)
+GO_LDFLAGS := -X $(VERSION_PACKAGE).version=$(VERSION)
 GO_LDFLAGS += -X $(VERSION_PACKAGE).buildDate=$(shell date +'%Y-%m-%dT%H:%M:%SZ')
 GO_LDFLAGS += -X $(VERSION_PACKAGE).commit=$(shell git rev-parse HEAD)
-GO_LDFLAGS +="
+
+CLI_LDFLAGS := -X $(CLICMD_PACKAGE).imagesB64=$(shell cat images.yaml | gzip | base64)
```
kars7e: NIT: could be `gzip -c images.yaml | base64`, one cat's life saved ;)

imikushin: No! This would kill the cat! 🙀

PR #510, opened 2018-06-13
 Support Tenancy in Dispatch

neosab:
The PR adds limited support for tenancy within Dispatch. Tenancy within Dispatch is represented using an `Organization` resource. What this means is that you have the ability to create multiple organizations and define policies to give access to users on those organizations. Users then can login to Dispatch by specifying the `organization` that they wish to access. This implementation is more like a soft-multitenancy and does not isolate the function execution environments within the underlying FaaS engine or define network policies for the function pods running on Kubernetes.

**Entity Store Changes**
* A new entity store method `ListGlobal` (open to suggestions w.r.t naming) to list entities across all organizations. This is needed by controllers & identity manager itself to list organization entities.
* Add a strict check to entity store methods for the presence of `organizationID` since it becomes really critical.
* LibKV store: The directory hierarchy has been changed from `<org_id>/<entity_type>/<entity_name>` `<entity_type>/<org_id>/<entity_name>`. Why? - with the former dir structure, you cannot use the existing List method to list all `Organization` entites as they are the TLD. With this flip, we can re-use the methods. This is similar to how K8S stores its data objects on etcd.  Note:- This does affect the value we store in the`key` column of PostgresSQL but it doesn't affect its uniqueness.

**Identity Manager Changes**
* Support for passing XDispatchOrg to policies & service account API's
* Enhance Casbin model to support organizationID
* Enhance auth methods (bearer/cookie) to support orgIDs.

TODO:
**API Manager** : Create unique api endpoints for each tenant.
**Event Driver**: Deploy unique pods for event drivers.
**Service Manager**: Fix ServiceClass etc to be multi-tenant aware.

WIP: Still pending some manual testing.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/510)


pkg/event-manager/drivers/entity_handler_test.go
```diff
@@ -58,6 +58,6 @@ func TestDriverDelete(t *testing.T) {
 	backend.On("Delete", mock.Anything, mock.Anything).Return(nil)
 	assert.NoError(t, handler.Delete(context.Background(), driver))
 	var drivers []*entities.Driver
-	es.List(context.Background(), "", entitystore.Options{}, drivers)
+	es.ListGlobal(context.Background(), entitystore.Options{}, drivers)
```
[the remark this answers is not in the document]
neosab: Fixed this, was an oversight in the test.

pkg/identity-manager/casbin_entity_adapter.go
```diff
@@ -37,15 +37,15 @@ func (a *CasbinEntityAdapter) LoadPolicy(model casbinModel.Model) error {
 	}
 
 	log.Debug("Reloading policies")
-	err := a.store.List(context.TODO(), IdentityManagerFlags.OrgID, opts, &policies)
+	err := a.store.ListGlobal(context.TODO(), opts, &policies)
```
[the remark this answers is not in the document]
neosab: Done

pkg/dispatchcli/cmd/manage_bootstrap.go
```diff
@@ -30,16 +30,17 @@ import (
...
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/vmware/dispatch/pkg/api/v1"
-	"github.com/vmware/dispatch/pkg/client"
-	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
+	"github.com/vmware/dispatch/pkg/identity-manager/gen/client/organization"
```
kars7e: something went wrong here, why is gen/client brought back to this file?

pkg/dispatchcli/cmd/manage_bootstrap.go
```diff
@@ -352,9 +359,9 @@ func runBootstrap(out, errOut io.Writer, cmd *cobra.Command, args []string) erro
 	policyName := defaultPolicyName
 	fmt.Fprintf(out, "Creating Policy: %s\n", policyName)
 	// Deleting any existing policy
-	_, err = iamClient.DeletePolicy(context.TODO(), policyName)
+	_, err = iamClient.DeletePolicy(context.TODO(), bootstrapOrg, policyName)
 	if err != nil {
-		if _, ok := err.(*client.ErrorNotFound); !ok {
+		if _, ok := err.(*policy.DeletePolicyNotFound); !ok {
```
kars7e: we should be using `client` errors

neosab: Yup, the result of a bad rebase, I fixed it after testing.

PR #513, opened 2018-06-13
Add reason field to function to hold function creation errors

### LINKED ISSUE #483: No helpful error output when giving bad arguments to function create (specifically --handler)
## Bug Report
I created a simple function (hello.py) and was trying to test out the source directory support:

```
$ tree hello
hello
└── hello.py
```

Where hello.py is the same as from the python examples.

I created a function as so:

```
dispatch create function hello-py-dir ./hello --handler hello.handler --image python3
```

Do you see the error?  I didn't.  The handler arg is wrong.  It should be "hello.handle".  I had an extra "r".

The result is a function which gets stuck for a while in CREATING and then finally errors out.  However there is nothing in the payload to indicate the cause of the error.  The only log with anything useful was the function pod log, but even that gave a cryptic python traceback.

### Expected behavior

Tell me know my handler is wrong. "Could not import function, is handler set correctly?"

### Impact

- [ ] High - Blocker

### Your Environment

rjew:
Fixes #483 

* Add `reason` field to function object
* When an error occurs during the function image build phase, return the last step that occurred (the one that failed) along with its output as the `reason`
    - It's difficult to determine the exact cause of the error since Docker combines both `stdout` and `stderr` into one stream
    - The only error Docker gives is the last command that failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. For more in
 ### Fixed
 
 - **Fix single file Java support.** Previously Java would fail for single files if no handler argument is provided. [PR #517](https://github.com/vmware/dispatch/pull/517)
+- [[Issue #483](https://github.com/vmware/dispatch/issues/483)] **No helpful error output when an error occurs during function create.**
+Added a `reason` field to the function object to hold function creation errors. [PR #513](https://github.com/vmware/dispatch/pull/513)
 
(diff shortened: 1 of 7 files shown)


rjew:
It does update the reason field. There was already a deferred call to `h.Store.UpdateWithError` in the function-manager controller.

imikushin:
An e2e-test of a failing function creation (e.g. with an invalid handler) would clearly demonstrate the reason field being populated

rjew:
Good point. All add an e2e test.

PR #516, opened 2018-06-14
Improve Dispatch errors in CLI & services

kars7e:
**Changes:**
- SDK now processes all swagger errors, and exports it's own error types.
- Adds identity manager client to SDK, and makes Dispatch CLI use it.
- Some errors were wrapped twice, causing the error message to be repeated. they should only be included once now.
- Improves consistency of returned errors between services:
  - Ensures that all handlers declare `401` and `403` errors.
  - Some services returned separate `500` error, while others only returned `default`. Now all returned `default` (which also includes `500` case).
  - Helper functions generating error messages for most common errors (404 Not Found, 409 Already Exists, 500 internal server) to ensure consistency between services.

**Note:**
All `*SwaggerError(err error) error` functions could be replaced by one function that switches over HTTP code of the response and extracts the message. I wasn't able to find a way to do it generated swagger client. If there is one, or if we could replace swagger client/change generated templates, these functions could go away (and with them many potential bugs).


pkg/client/functions.go
```diff
@@ -127,6 +174,23 @@ func (c *DefaultFunctionsClient) ListFunctionRuns(ctx context.Context, organizat
 	return runs, nil
 }
 
+func listRunsSwaggerError(err error) error {
+	if err == nil {
+		return nil
+	}
+	switch v := err.(type) {
+	case *runner.GetRunsBadRequest:
+		return NewErrorUnauthorized(v.Payload)
```
tenczar: You seem to be missing the BadRequest here.

kars7e:
I agree, and I will keep searching for a way to remove all this boilerplate.

PR #517, opened 2018-06-14
Fix single file java support

rjew:
Don't write root directory into the tar so now when a single file is used only a single file is copied (previously both the single source file and tmp root directory were written causing issues for Java)


pkg/utils/tar.go
```diff
@@ -46,6 +46,12 @@ func Tar(source string, w io.Writer) error {
 			if err != nil {
 				return err
 			}
+
+			// Skip writing tar header for root dir
+			if info.IsDir() && path == source {
+				return nil
+			}
```
imikushin: I'd add something like this to `TestWriteSourceDir` (in builder_test.go) to test against what this code is fixing:
```go
	ls, err := ioutil.ReadDir(tmpDir)
	require.NoError(t, err)
	assert.Equal(t, 1, len(ls))
	assert.Equal(t, "mypkg", ls[0].Name())
```
The test should fail without the change and pass with it :)

PR #527, opened 2018-06-19
Add model field pattern

zimengyang:
* Change all model field pattern, name should only contain undersocre, dash, letter or number and starts with alphabetical characters  
* Translate regex pattern to user-friendly msg 
* Add name validation for eventdriver

```
$ dispatch create secret my.secret ./secret.json
[Code: 605] name should start with letter or number and may only contain letters, numbers, underscores and dashes
```


pkg/api/v1/field_patterns.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// regex pattern
+	Pattern string `json:"pattern"`
+
+	// translated msg presented to user
+	Message string `json:"message"`
+}
+
+// FieldPatternWordNumberDashOnly letter, underscore, number and dash
+var FieldPatternWordNumberDashOnly = FieldPattern{
```
tenczar: I'm not sure this name is entirely helpful. `FieldPatternWordNumberDashOnly` and `FieldPatternLetterNumberDashOnly` differ at `Letter` and `Word, where the actual difference is one accepts underscores and the other doesn't.

pkg/api/v1/field_patterns.go
```diff
@@ -0,0 +1,44 @@
...
+// FieldPatternLetterNumberDashOnly letter, number and dash
+var FieldPatternLetterNumberDashOnly = FieldPattern{
+	Pattern: `^[a-zA-Z0-9][a-zA-Z0-9\-]*$`,
+	Message: " should start with letter or number and may only contain letters, numbers and dashes",
+}
+
+// Validate validates field pattern and return error with translated msg
+func (p FieldPattern) Validate(fieldName, field string) error {
+	if err := validate.Pattern(fieldName, "body", field, p.Pattern); err != nil {
+		return errors.New(err.Code(), fieldName+p.Message)
```
tenczar: won't this make the error message look like, for example, `nameshould start with ...`?

pkg/api/v1/field_patterns.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// FieldPatternWordNumberDashOnly letter, underscore, number and dash
+var FieldPatternWordNumberDashOnly = FieldPattern{
+	Pattern: `^[\w\d][\w\d\-]*$`,
+	Message: " should start with letter or number and may only contain letters, numbers, underscores and dashes",
+}
+
+// FieldPatternLetterNumberDashOnly letter, number and dash
+var FieldPatternLetterNumberDashOnly = FieldPattern{
```
tenczar: Do we ever use this?

zimengyang: For now, this one was only used in EventDriver name.

pkg/api/v1/field_patterns.go
zimengyang: A space is in the beginning of message. But this is not good, I will change this to Join(). Thanks!

pkg/api/v1/field_patterns.go
zimengyang: I agree this is somehow confusing. The LetterNumberDashOnly one is only used in EventDriver. The rest of model name will use WordNumberDashOnly. What do you think if I change them to `FieldPatternName` and `FieldPatternNameNoUnderscore`?

tenczar: These seem reasonable.

zimengyang: Thanks, will do!

PR #536, opened 2018-06-21
fix service manager

berndtj:
Service manager was broken due to the changes with regards to passing
org IDs around.

Extened the timeouts to try and address the e2e test failures.

### DIFF
--- a/charts/dispatch/charts/service-manager/templates/deployment.yaml
+++ b/charts/dispatch/charts/service-manager/templates/deployment.yaml
@@ -33,7 +33,6 @@ spec:
             - "--tls-port=443"
             - "--tls-certificate=/data/tls/tls.crt"
             - "--tls-key=/data/tls/tls.key"
-            - "--organization=$(ORGANIZATION)"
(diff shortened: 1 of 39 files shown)


kars7e:
Looks like there is an issue with one of the mocks, unit tests don't pass

berndtj:
Ugh... I need more logs.  The tests work for me.

berndtj:
Ugh... confused how the last patch passed e2e with the above failures... should be a relatively easy fix though.

pkg/client/secrets.go
```diff
@@ -53,6 +54,7 @@ func (c *DefaultSecretsClient) CreateSecret(ctx context.Context, organizationID
 		XDispatchOrg: c.getOrgID(organizationID),
 		Secret:       secret,
 	}
+	log.Infof("creating secret %s/%s", organizationID, *secret.Name)
```
kars7e: Some debug leftovers?

berndtj: I decided to leave it in... I can remove

berndtj: done

PR #537, opened 2018-06-21
Add follow and last flags to get runs 

### LINKED ISSUE #465: Tail/follow function runs as they happen (or at least get last run)
## Feature Request
`dispatch get runs` works fine, but isn't very easy to use.  The user needs to give the name of the function and the UUID associated with the actual instance run that they care about.  When developing functions or workflows most folks just want to view the stream of runs like a log.  This is a proposal for the following:

`dispatch get runs [function name] --follow`

Which would follow all runs as they happen (filtered by function name if present)

`dispatch get runs [function name] --last`

Which would output the last run (chronologically)

### Complexity

- [x] Low - Simple enhancement or bug fix, no architectural changes or refactoring

rjew:
Fixes #465 

* Added `--follow, -f` and `--last` flags to get runs for tailing runs and getting the last executed run respectively
* Added `since` query param to get runs API
* Added unit tests to `handler` and `controller` to ensure runs are correctly gotten and deleted
* Added examples for retrieving runs

**Note:**
When following runs there is a possibility that the same run could be printed out twice. There is a small window between the ticker and making the request where if an execution falls within this window, it will be repeated on the next tick. 

There doesn't seem to be a way to eliminate this except by putting the new ticker time after the request but that introduces a new window for missing runs.


pkg/client/functions.go
```diff
@@ -140,10 +141,11 @@ func getRunSwaggerError(err error) error {
 }
 
 // ListRuns lists all the available results from previous function runs
-func (c *DefaultFunctionsClient) ListRuns(ctx context.Context, organizationID string) ([]v1.Run, error) {
+func (c *DefaultFunctionsClient) ListRuns(ctx context.Context, organizationID string, since *int64) ([]v1.Run, error) {
```
kars7e: Does it have to be a pointer? would 0 be an ok default value? Also it seems `time.Duration` or `time.Time` would be more appropriate here.

pkg/client/functions.go
```diff
@@ -23,9 +23,9 @@ import (
 type FunctionsClient interface {
 	// Function Runner
 	RunFunction(ctx context.Context, organizationID string, run *v1.Run) (*v1.Run, error)
-	GetFunctionRun(ctx context.Context, organizationID string, functionName string, runName string) (*v1.Run, error)
-	ListRuns(ctx context.Context, organizationID string) ([]v1.Run, error)
-	ListFunctionRuns(ctx context.Context, organizationID string, functionName string) ([]v1.Run, error)
+	GetFunctionRun(ctx context.Context, organizationID string, functionName string, runName string, since *int64) (*v1.Run, error)
```
kars7e: functions are growing in arguments, and I expect there will be more. It might be a good time to introduce `FunctionOpts` structure.

pkg/dispatchcli/cmd/get_run.go
```diff
@@ -58,35 +81,130 @@ func getFunctionRun(out, errOut io.Writer, cmd *cobra.Command, args []string, c
...
+	since := time.Now()
+	resp, err := c.ListRuns(context.TODO(), "", nil)
 
 	if err != nil {
 		return err
 	}
-	return formatRunOutput(out, true, resp)
+	if last && len(resp) > 0 {
+		lastRun := resp[0]
+		for _, run := range resp {
```
kars7e: This is ok for now, but it seems like something that should be done server-side, i.e. sort by execution time and then retrieve the first result. I would imagine the number of runs for single will grow considerably, and some pagination/filtering on the server side will need to be added anyway (like you did with "since")

rjew: Yeah the number of runs could definitely get considerably large so this is more of a temp solution

berndtj: +1

Issue #525, opened 2018-06-18
Eventdriver name should NOT allow underscore


zimengyang:
### Current behavior
Eventdriver with a underscore in name will case k8s error:
```
Deployment.apps "event-driver_test" is invalid: metadata.name: Invalid value: "driver_test":a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'
```

### Steps to reproduce
```
dispatch create eventdriver aws --name event-driver_test
```
### Impact
- [X] Low - Annoyance, but does not impact business or functionality
- [ ] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

### Possible solution
Add name pattern validation 

### Your Environment

* Dispatch CLI version (or git commit): 0.0.17
* Dispatch Chart version (or image[s] tag):
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

Issue #523, opened 2018-06-16
Event manager throws 500 when creating an event driver with . (dot) in the name


kars7e:
### Expected behavior
When creating an event driver with `.` in name, event manager should return a meaningful message that dots are not supported

### Current behavior
Currently, it throws a 500 error with a meaningless message. The underlying error is coming from entity store, which is not accepting the entity with dotted name, and the error bubbles to the response as 500.

Other services have a pattern defined for `name` attribute in model definitions. For other services, the error returned is (example for secret):
```
[Code: 605] create Secret error: name in body should match '^[\w\d\-]+$'
```
Note that this error requires users to parse the regular expression which is not ideal, but it's better than 500.

### Impact

- [ ] Low - Annoyance, but does not impact business or functionality
- [x] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

### Possible solution
Add validation pattern to event-manager `event-driver` model definition.

### Your Environment

* Dispatch CLI version (or git commit):
* Dispatch Chart version (or image[s] tag): 0.1.17
* Operating System and version:
* Kubernetes version (and distribution):
* Etc (any other useful environmental information):

PR #540, opened 2018-06-25
fix eventdriver name filter: lower letter only

zimengyang:
Eventdriver name filter: lower case only
Add e2e test for invalid names


zimengyang:
Fix #525 and #523

PR #548, opened 2018-06-27
Move source out of function entity and into its own entity store

rjew:
* `Source` is now its own entity in the store, where `Function` now references the source by its entity name
* Add `sourceEntityHandler` in controller to handle source entities that need to be resolved such as when a source is added but its corresponding function fails to be added
* Add `Source` api model


kars7e:
Can you add some rationale for this change? Looking at the code, it's not immediately obvious to me why is it needed.

rjew:
This is an initial change to move the source out of the function entity since we don't currently have any limit on the size of the source. This will reduce the load when retrieving the function when the source is large and will also help if we plan to store the source somewhere else in the future or if we want to support specifying functions using s3 or github.

kars7e:
Thanks, this is helpful. A thought I had after reading this - the fact that the source has its own name makes me again think that using names is primary key isn't the best idea (but this is, of course, unrelated to this change)

pkg/api/v1/source.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Source source
+// swagger:model Source
+type Source struct {
+
+	// code
+	Code strfmt.Base64 `json:"code,omitempty"`
+
+	// the name of the function associated with
+	// Required: true
```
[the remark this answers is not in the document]
rjew: Yeah good point it does seem a bit redundant. Probably don't need them in the API model and just have the `Name` field in the entity be an ID like you and @kars7e mentioned.

pkg/api/v1/source.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/swag"
+)
+
+// NO TESTS
+
+// Source source
+// swagger:model Source
+type Source struct {
```
berndtj: We talked off-line and I assume that the URL is actually going in the Function object to point to the Source entity which is just the code?

pkg/function-manager/controller.go
```diff
@@ -27,6 +27,67 @@ type ControllerConfig struct {
 	ResyncPeriod time.Duration
 }
 
+type sourceEntityHandler struct {
```
berndtj: Do we need a separate control loop for the source?  We aren't resolving any state.  I could certainly see an interface to source, but I think the "control loop" may not be the right one.

rjew: Right now the control loop is used to resolved orphaned source entities, i.e. when a source entity is added to the store but the function entity fails to be added, the source status is set to `DELETING` and updated which the controller will pick up and delete. 

The control loop could be eliminated if a hard delete is used instead when a source becomes orphaned.

PR #550, opened 2018-07-02
Create API paths under org name

neosab:
Creates the user-specified API paths under the org name. This is because we share an API Gateway and the paths from different org's can overlap. For production uses, user usually specifies a hostname for the API's.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/550)


neosab:
@berndtj can you reapprove when you get a chance? Had to fix a minor issue.

PR #551, opened 2018-07-02
Add CI base images pipeline

zimengyang:
* Adding a new pipeline in CI, which is for testing language base images under dispatchframework repo. 
* Update CI README

In this pipeline, some key jobs are:
* `create/delete gke cluster` jobs are triggered manually
* `deploy/update dispatch` job is triggered by Dispatch release
* `build base image` job is trigger by base image repo pr, and then `test base image` job will be triggered using new image. So each language base image will have separate build & test jobs.

A testing pipeline has been setup in ci.dispatchframework.io.

**TODO**:
* `reuse scripts` - some scripts and task files are partially shared across pipelines, which might be able to simplify and extract separated script. It will need some refactor
* `tests` - add tests for each language base image.
* UI layout is somehow messy. Haven't found a way to  change it.

### DIFF
--- a/ci/pipelines/base-images.yml
+++ b/ci/pipelines/base-images.yml
@@ -1,0 +1,542 @@
+---
+resource_types:
+
+- name: keyval
+  type: docker-image
+  source:
+    repository: swce/keyval-resource
+
+- name: pull-request
+  type: docker-image
+  source:
+    repository: jtarchie/pr
+
+
+resources:
+
+# dispatch repo
+- name: dispatch-master
+  type: git
+  source:
+    uri: https://github.com/vmware/dispatch.git
+    branch: master
+
+# latest release contains binaries
+- name: dispatch-release
+  type: github-release
+  source:
+    owner: vmware
+    repository: dispatch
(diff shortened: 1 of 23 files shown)


kars7e:
LGTM so far!

ci/pipelines/base-images.yml
```diff
@@ -0,0 +1,542 @@
...
+  - task: deploy-dispatch
+    file: dispatch/ci/base-images/install-dispatch.yml
+    params:
+      GKE_KEY: ((gke-key))
+      GKE_PROJECT_ID: ((gke-project-id))
+      DOCKER_REGISTRY_HOST: ((docker-hub-org))
+      FAAS: openfaas
+      EVENT_TRANSPORT: rabbitmq
+    input_mapping:
+      version: version
```
kars7e: NIT: this shouldn't be needed if the name is the same, did you find it not working without the mapping?

kars7e: also since you are getting the dispatch-resource, can't you extract version from it, is the version resource needed?

PR #563, opened 2018-07-10
Add bash completion script

ali5ter:
It might be useful for the dispatch CLI to generate autocompletion scripts for bash, zsh, etc. similarly to kubectl. In the meantime however, this bash completion script might be useful.


ali5ter:
@berndtj generated using cli_taxo.py script from https://github.com/ali5ter/cli_taxo. The invocation is:

cli_taxo.py dispatch \
    --commands-token '^Available Commands:' \
    --commands-filter '^\s\s([\w|-]+?)\s\s+' \
    --options-token '^Global Flags:|^Flags:' \
     -o bash -O

Here's example of how it could be automated to generate completion file... https://github.com/ali5ter/vmware_scripts/blob/master/vke/create_completion

PR #565, opened 2018-07-10
Add e2e tests to release pipeline to test upgrade to new release

### LINKED ISSUE #419: Release job should test upgrade from previous to new release
## Bug Report
There is already an issue to run e2e on a new release (not currently done), to ensure that each release is functional.  We should also test upgrade from one release to the next.  This doesn't initially have to be blocking, but we should know and notify users of any potential breakage or incompatibilty

### Expected behavior

1. run e2e on current release
2. upgrade to new release
3. validate e2e on new release

### Current behavior

N/A

### Impact

- [ ] High - Blocker

### Your Environment

rjew:
Fixes #419 

* Adds new upgrade e2e tests to the release pipeline
* This is currently non-blocking but can easily be added in the future to block the release if the upgrade e2e tests fail
* Upgrade e2e tests are separated into two phases:
    * `Setup` - Run in the old dispatch release, creates the resources needed to test in the new dispatch release
        * `Setup.bats` loads the resources in a specific order since the tests are not independent of each other
    * `Exec` - Run in the new dispatch release, acts on the resources created in the `Setup` phase to ensure that all the old resources can still be acted on correctly (i.e. functions can still be executed, resources can still be deleted, etc.)
        * Tests in the `Exec` phase do not affect another test's resources so they can be run in any order

### DIFF
--- a/ci/e2e/deploy-dispatch.yml
+++ b/ci/e2e/deploy-dispatch.yml
@@ -13,6 +13,8 @@ params:
   DOCKER_REGISTRY_HOST:
   FAAS: openfaas
   EVENT_TRANSPORT: kafka
+  BOOTSTRAP: 1
+  CI_KEYS_TEST: 0
 
 # dispatch must be dispatch git repo.
 # dispatch-cli must contain "dispatch" binary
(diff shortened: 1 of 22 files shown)


kars7e:
If I got #419 right, The intent was to deploy old release first, run some operations, and then, using the same GKE cluster, deploy the new version of dispatch and verify that everything works correctly. Can you confirm @berndtj

berndtj:
@kars7e is correct.  We want to test that a user with an existing dispatch deployment can safely upgrade to the next release without breakage.  Or at least we want to know if there will be breakage (pre-GA).

kars7e:
Closing outdated PRs

PR #571, opened 2018-07-16
Optimizing image pulls in image manager

neosab:
**Current Behavior**
* We pull `BaseImage`s every time it's missing from the docker host.
* We build and push `Images`s every time it's missing from the docker host.

This is not needed and causes a surge of activity that hogs cpu/mem/network whenever you kill an image-manager pod and a new one is started.

**New Behavior**
* Pull`BaseImage`s/'`Images`s only if it's missing and a certain time has elapsed since last successful pull.
* We will stop building and pushing `Image` if the image is just in missing state. Instead, a pull would suffice to verify that the container image exists on the registry.

**TODO**
Prune images/base-images from docker host based on some cache eviction strategy. This will ensure that we don't eventually pull all images/base-images on the docker host.


pkg/image-manager/mocks/common_api_client.go
```diff
@@ -0,0 +1,2046 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
kars7e: there is already a mock for docker client in `pkg/functions/docker/mocks`, maybe we could move mocks of external services into a top-level mocks directory and re-use them

pkg/image-manager/controller.go
```diff
@@ -114,7 +116,13 @@ func (h *imageEntityHandler) Add(ctx context.Context, obj entitystore.Entity) (e
 
 	defer func() { h.Store.UpdateWithError(ctx, i, err) }()
 
-	if err = h.Builder.imageCreate(ctx, i, &bi); err != nil {
+	if i.Status == entitystore.StatusMISSING {
```
kars7e: `dispatch-server` does not use external registry by default, seems like this will break it. Can we make this behavior configurable through an option?

kars7e: Actually, we should try to pull in every case, and if the error is that image does not exist in the registry, we should re-build it. This will also cover the case where image was deleted both from docker & registry.

pkg/image-manager/mocks/common_api_client.go
neosab: Yeah, we could. I will move them.

pkg/image-manager/controller.go
neosab: I think this is what happens i.e if the image is `MISSING` from docker host, we try to pull it and if it does not exists in the Registry as well (or some other error occurred), the image is then moved to the `ERROR` state. This would result in the image getting rebuilt.

The difference is that we do this check if the image exists at a longer interval than before. So if an image is removed from the registry, it will eventually be rebuilt.

kars7e: LGTM

PR #573, opened 2018-07-17
Nkaashoek/ZookeeperScaling

nickaashoek:
* Added code to contact zookeeper as a sort of lock server

* Added configuration information needed to find zookeeper

* Build our own zookeeper image


lib/riff/requester.go
```diff
@@ -136,18 +148,47 @@ func (r Requester) makeHeaders(runID string) message.Headers {
...
-	case msg := <-resultChan:
-		return msg.Payload(), nil
+	case e := <-events:
+		if e.Type == zk.EventNodeCreated {
+			payload, _, err := client.Get(runNode)
+			if err != nil {
+				clearZnode(runNode, client)
+				return nil, errors.Errorf("Unable to get payload from znode for run %v: %v", reqID, err)
+			}
+			clearZnode(runNode, client)
```
tenczar: Why not just call `clearZnode` before the err handling block?

pkg/function-manager/controller.go
```diff
@@ -244,6 +249,9 @@ func (h *runEntityHandler) Add(ctx context.Context, obj entitystore.Entity) (err
 		Secrets:  run.Secrets,
 		Services: run.Services,
 	}, run.Input)
+
+	log.Infof("Received results of runner: \n\t Output: %v \n\t Err: %v", output, err)
```
tenczar: This might leak sensitive data to the logs.

lib/riff/requester.go
```diff
@@ -87,6 +91,19 @@ func (r *Requester) run() {
...
+	client, _, err := zk.Connect([]string{r.zookeeperLocation}, time.Second)
+	if err != nil {
+		log.Errorf("Unable to connect to zookeeper: ", err)
+	}
+	defer client.Close()
+
+	if exists, _, _ := client.Exists("/riffRuns"); !exists {
+		_, err := client.Create("/riffRuns", []byte{}, int32(0), zk.WorldACL(zk.PermAll))
+		if err != nil {
+			log.Fatalf("Unable to create riffRuns node: %v", err)
```
kars7e: log fatal will kill the process. This is probably ok for now (we can't operate without zk working), but imho some retry with logging should be applied later. eventually, we should include things like that in the readiness reported to kubernetes.

pkg/dispatchcli/cmd/install.go
```diff
@@ -1088,6 +1101,7 @@ func runInstall(out, errOut io.Writer, cmd *cobra.Command, args []string) error
 			"global.registry.uri":                                 config.DispatchConfig.ImageRegistry.Name,
 			"global.registry.insecure":                            strconv.FormatBool(config.DispatchConfig.ImageRegistry.Insecure),
 			"global.tls.secretName":                               config.DispatchConfig.TLS.SecretName,
+			"global.zookeeper.location":                           config.Zookeeper.Location,
```
neosab: `config.Zookeeper.Location` needs to be changed when installing in the single namespace mode. This is similar to what we did for kafka brokers in https://github.com/vmware/dispatch/blob/9120bdd5b43cc6eebdfd9c6a23a67e95e986ebbb/pkg/dispatchcli/cmd/install.go#L794

pkg/dispatchcli/cmd/install_config.go
```diff
@@ -109,6 +109,14 @@ jaeger:
       cassandra.resources.requests.memory: 2Gi
   agent:
   enabled: false
+zookeeper:
+  chart:
+    chart: zookeeper
+    namespace: zookeeper
+    release: zookeeper
+    repo: http://storage.googleapis.com/kubernetes-charts-incubator
+    version: 1.1.0
```
neosab: Do we need to pin the version?

pkg/controller/controller.go
```diff
@@ -90,11 +94,22 @@ func NewController(options Options) Controller {
...
+		options.ZookeeperLocation = "127.0.0.1"
+	}
+	if options.Driver == nil {
+		driver, err := zookeeper.NewDriver(options.ZookeeperLocation)
+		if err != nil {
+			log.Fatalf("Unable to get zookeeper driver for controller")
+		}
+		options.Driver = driver
+	}
+	log.Infof("Connected to zookeeper at address %v", options.ZookeeperLocation)
```
neosab: Nit: move this into the `if` block where the driver is instantiated.

pkg/controller/controller.go
```diff
@@ -233,15 +248,27 @@ func (dc *DefaultController) run(stopChan <-chan bool) {
...
+	defer dc.driver.Close()
+
+	if err := dc.driver.CreateNode("/entities", []byte{}); err != nil {
+		log.Fatalf("Unable to create overarching znode %v", err)
+	}
 	// Start a worker pool.  The pool scales up to dc.options.Workers.
 	go func() {
 		sem := semaphore.NewWeighted(int64(dc.options.Workers))
 		for watchEvent := range dc.watcher {
+			lock, canModify := dc.driver.LockEntity(watchEvent.Entity.GetName())
```
neosab: Shouldn't we first acquire the semaphore and then lock the entity? This way we can let another controller process the entity if this controller is out of workers.

neosab: Also, the entity name is not unique across different entity types and also across organizations. We should probably use `GetID`.

pkg/dispatchcli/cmd/install.go
```diff
@@ -962,6 +968,13 @@ func runInstall(out, errOut io.Writer, cmd *cobra.Command, args []string) error
 		}
 	}
 
+	if installService("zookeeper") {
```
neosab: I don't see an uninstall for zookeeper, did I miss it?

PR #581, opened 2018-07-26
Ability to set cpu and memory limits/requests on a function

### LINKED ISSUE #301: Set cpu/memory limits on a function
- [ ] Bug
- [x] Feature
- [ ] Enhancement

## Detailed Description
It would be nice if we are able to set cpu and memory limits on a function in order to have better control on what resources are being used from it when it is triggered. Maybe it would be good if these limits are introduced, to have some default values if they are not provided when a function is created.

rjew:
Fixes #301

* Add configurable cpu and memory limits/requests for a function
* Add cpu and memory limits/requests to riff and kubeless
* Add FunctionResources model
* Add unit tests and e2e test

### DIFF
--- a/pkg/functions/kubeless/driver.go
+++ b/pkg/functions/kubeless/driver.go
@@ -22,9 +22,10 @@ import (
 	"github.com/vmware/dispatch/pkg/functions"
 	"github.com/vmware/dispatch/pkg/trace"
 	"github.com/vmware/dispatch/pkg/utils"
-	"k8s.io/api/core/v1"
+	kapi "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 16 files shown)


pkg/functions/kubeless/driver.go
```diff
@@ -120,17 +140,75 @@ func (d *kubelessDriver) Create(ctx context.Context, f *functions.Function) erro
...
+		resourceRequirements.Limits[kapi.ResourceMemory] = qty
+	}
+	if funcRequests.CPU != "" {
+		qty, err := resource.ParseQuantity(funcRequests.CPU)
+		if err != nil {
+			return errors.Wrapf(err, "error parsing cpu request '%s'", funcRequests.CPU)
+		}
+		resourceRequirements.Requests[kapi.ResourceCPU] = qty
+	}
+	if funcRequests.Memory != "" {
```
tenczar: Seeing this chunk of code that is essentially the same as the code in NewRiffTalk makes me think that this really needs it's own object to fully handle populating the kapi.ResourceRequirements object. or at the very least this functionality should be moved into the FunctionResources object and then shared here and in the NewRiffTalk method.

rjew: Sharing code between NewRiffTalk and the Kubeless driver is not really feasible since we currently vendor the riff library with its own version of k8s dependencies. This leads to conflicts when passing around k8s objects, so the resource requirement has to be parsed into the correct version of `kapi.Quantity` in riffTalk.

kars7e:
Closing outdated PRs

PR #586, opened 2018-08-02
[CLI] Add set,delete,current for manage context

zimengyang:
Mange dispatch contexts by `dispatch context`

* Add set,delete and current for managing context
```bash
$ dispatch context current
$ dispatch context set <CONTEXT>
$ dispatch context delete <CONTEXT>
```
* Listing all contexts remains as `dispatch context`

### DIFF
--- a/pkg/dispatchcli/cmd/manage_context.go
+++ b/pkg/dispatchcli/cmd/manage_context.go
@@ -49,6 +49,7 @@ func NewCmdManageContext(out io.Writer, errOut io.Writer) *cobra.Command {
 		Long:    manageContextLong,
 		Example: manageContextExample,
 		Aliases: []string{"app"},
+		Args:    cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return formatContextOutput(out)
 		},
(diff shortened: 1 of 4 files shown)


pkg/dispatchcli/cmd/manage_context.go
```diff
@@ -21,48 +21,81 @@ import (
...
+	return nil
+}
+
+func deleteContext(out, errOut io.Writer, cmd *cobra.Command, args []string) error {
+	ctxName := args[0]
+	if _, ok := cmdConfig.Contexts[ctxName]; !ok {
+		return errors.Errorf("No such context %s", ctxName)
+	}
+
+	if cmdConfig.Current == ctxName {
```
neosab: What if there is no other context?

Anyway, I would prefer to delete the current active context. We can then display an error message saying no context has been set every time user executes a command. Also, ensure the error is only thrown when the config file is present and the context is nil because we do have usecase like install/single-binary which you can run without a config.json file.

zimengyang: Delete active context is enabled. For now, viper will try to load config file, if current context was deleted from contexts, cli will try to read values from flags. The 'current context was found' msg will be displayed using 'context current' cmd. So other use cases should not be affected.

neosab:
`dispatch manage context set CONTEXT` seems too long to me. Should we move context to a subcommand to dispatch? IMO, `manage` operates more on the dispatch installation but `context` is just a client configuration.

zimengyang:
Yes, codes of `manage context` were removed. @berndtj

PR #591, opened 2018-08-18
Knative based function-manager

imikushin:
Implemented Functions and Secrets API on top of Knative. Updated the CLI accordingly. 
Fixed nits found by Go v1.11 compiler.


imikushin:
Not storing function run info for now. 

We need to redesign this bit, taking into account concerns about payload data privacy and storing / processing the invocation data.

pkg/function-manager/backend/knhelpers_test.go
```diff
@@ -0,0 +1,6 @@
+///////////////////////////////////////////////////////////////////////
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+///////////////////////////////////////////////////////////////////////
+
+package backend_test
```
tenczar: Is this necessary?

pkg/function-manager/knhandlers_test.go
```diff
@@ -0,0 +1,6 @@
+///////////////////////////////////////////////////////////////////////
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+///////////////////////////////////////////////////////////////////////
+
+package functionmanager
```
tenczar: Maybe add a TODO here so that Zimeng or I remember to pick this up.

PR #592, opened 2018-08-20
Changed service name to match upstream change

nickaashoek:


neosab:
Cool! Let's wait for E2e

Issue #594, opened 2018-08-21
Support url's in CLI create -f


neosab:
## Feature Request
Would be awesome if we support URL's in the create CLI e.g
dispatch create -f https://raw.githubusercontent.com/vmware/dispatch/master/examples/seed.yaml

Currently, this needs to be a file

### Complexity

- [X] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

PR #597, opened 2018-08-21
Add create image by url

pzmrzy:
Fix Issue (#594 )

### DIFF
--- a/pkg/dispatchcli/cmd/create.go
+++ b/pkg/dispatchcli/cmd/create.go
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
+	"os"
 	"path"
 	"path/filepath"
 	"strings"
@@ -31,6 +33,8 @@ var (
 	createExample = i18n.T(``)
 	file          = i18n.T(``)
 	workDir       = i18n.T(``)
+	baseURL       = i18n.T(``)
+	isURL         = false
 )
 
 // ModelAction is the function type for CLI actions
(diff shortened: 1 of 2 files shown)


pkg/dispatchcli/cmd/create.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/vmware/dispatch/pkg/api/v1"
 	"github.com/vmware/dispatch/pkg/dispatchcli/i18n"
 	"github.com/vmware/dispatch/pkg/utils"
+	"os"
```
neosab: Std Library imports go to the first block

Some pointers : https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/dispatchcli/cmd/create.go
```diff
@@ -31,6 +33,7 @@ var (
 	createExample = i18n.T(``)
 	file          = i18n.T(``)
 	workDir       = i18n.T(``)
+	tmpPath      = i18n.T(`/tmp/seed.yaml`)
```
neosab: Don't hardcode the tmpPath, use `ioutil.TempFile` and defer its removal.

pkg/dispatchcli/cmd/create.go
```diff
@@ -63,6 +66,39 @@ func importFile(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []stri
 	return importBytes(out, b, actionMap, actionName)
 }
 
+func importFileWithURL(out io.Writer, errOut io.Writer, cmd *cobra.Command, args []string, actionMap map[string]ModelAction, actionName string) error {
+	err := downloadFile(file, tmpPath)
```
neosab: Actually, you don't even have to download the file to tmp dir, you could do this in-memory by reading the response and passing to `importBytes`

pkg/dispatchcli/cmd/create.go
```diff
@@ -9,16 +9,16 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/ghodss/yaml"
```
neosab: This is not stdlib, you should move it back.

pkg/dispatchcli/cmd/create.go
```diff
@@ -9,16 +9,16 @@ import (
...
+	"github.com/pkg/errors"
```
neosab: same

pkg/dispatchcli/cmd/create.go
```diff
@@ -9,16 +9,16 @@ import (
...
+	"github.com/spf13/cobra"
```
neosab: same here

berndtj:
Is there an issue linked to this?  Please squash commits and add helpful message.

berndtj:
Also, can we add an e2e for this?

neosab:
Another improvement could be the resolution of Function `sourcePath`  which currently looks at the local filesystem. It will be nice if the CLI could try the local path else do a relative URL path while operating in this mode. I just realized the `seed.yaml` still won't work after this addition.

neosab:
@pzmrzy Please add an E2E test for this. The test should use the `seed.yaml` that's part of the PR being tested by the CI job, otherwise, the `yaml` being tested may not be in sync with the API changes in that PR. This could lead to unnecessary CI failures.

PR #601, opened 2018-08-22
Update K8S Version to 1.9.6-gke.2 in E2E CI

neosab:
Currently our CI jobs are failing to create a K8S cluster since the previous version `1.9.6-gke.1` is unsupported for Master Nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/dispatch/601)

### DIFF
--- a/ci/e2e/gke-cluster-create.yml
+++ b/ci/e2e/gke-cluster-create.yml
@@ -10,7 +10,7 @@ image_resource:
 params:
   GKE_KEY:
   GKE_PROJECT_ID:
-  K8S_VERSION: 1.9.6-gke.1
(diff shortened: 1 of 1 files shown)


ci/e2e/gke-cluster-create.yml
```diff
@@ -10,7 +10,7 @@ image_resource:
 params:
   GKE_KEY:
   GKE_PROJECT_ID:
-  K8S_VERSION: 1.9.6-gke.1
+  K8S_VERSION: 1.9.6-gke.2
```
imikushin: why not `1.10.6-gke.1`?

PR #604, opened 2018-08-22
Dispatch apply

pzmrzy:


zimengyang:
In my opinion, the functionality of `apply` looks pretty overlapped with `update`. `apply` will create new entity instead of throwing error.  What about adding a flag in `update` to create a new entity if no one present? (like `helm upgrade --install RELEASE`)

berndtj:
I agree I had thought this would be addressed with a flag to update (i.e. update -i).  Also, please squash commits and provide a good commit message.

neosab:
Yeah, agree to have one unified command. 

I would rather follow `kubectl` than `helm` CLI approach (as we have done before in our CLI's). How about just one `apply` command that updates/creates new resource. 

I don't think the current update behavior of throwing an error when a resource doesn't exist is that useful. Say I have an all-in-one yaml and I add new resources to it e.g. add a new event subscription to an existing function, I would expect this to just work with whatever update command that's provided.

neosab:
As discussed, proceed with `update -i` instead of `apply`. We can add `apply` if needed later.

kars7e:
I was able to convince Berndt to `apply` :) Rationale: the command used more often should be the top level one, and I don't see us using `update` anymore if `apply` semantics are available (apart from specific use cases). We use `kubectl apply` everyday, but `kubectl patch` is very rare (although it's not an exact equivalent of `dispatch update`, it's close enough).

PR #605, opened 2018-08-22
[WIP] Moving API Manager to Knative/Istio

nickaashoek:
Goals:
- Get rid of async
- Use istio to manager our APIs

TODO:
- Authentication/HTTPs
- Fully support all features of istio virtual services
- Install using KO


pkg/api-manager/istiohandler.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	return endpoint.NewGetApisOK().WithPayload(results)
+}
+
+func (cl *IstioHandlers) UpdateAPI(params endpoint.UpdateAPIParams, principal interface{}) middleware.Responder {
+	span, _ := trace.Trace(params.HTTPRequest.Context(), "")
+	defer span.Finish()
+
+	log.Infof("Trying to update: %+v", params)
+	return cl.AddAPI(endpoint.AddAPIParams{
```
berndtj: Does update work?  Do we need to delete the endpoint first?

pkg/api-manager/istio/istio.go
```diff
@@ -0,0 +1,91 @@
...
+	crdClient, err := crd.NewClient("", "", []model.ProtoSchema{model.VirtualService}, "")
+	if err != nil {
+		return nil, err
+	}
+	return &Client{
+		istioClient: crdClient,
+	}, nil
+}
+
+type Client struct {
```
imikushin: You don't really need another layer of istio client abstraction: there's already an API client that you're using here in the `istioClient` field.

pkg/api-manager/istiohandler.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func (cl *IstioHandlers) AddAPI(params endpoint.AddAPIParams, principal interface{}) middleware.Responder {
+	span, ctx := trace.Trace(params.HTTPRequest.Context(), "")
+	defer span.Finish()
+
+	log.Infof("Trying to add istio api: %+v", params)
+
+	api := apiModelOntoIstioEntity(params.Body)
+	virtualServiceSpec, err := yaml.Marshal(api)
```
imikushin: Separate Istio data model and YAML marshaling/unmarshaling seems redundant: why not just use istioClient directly?

pkg/api-manager/types.go
```diff
@@ -0,0 +1,52 @@
+package apimanager
+
+type VirtualService struct {
```
imikushin: Looks like these are redundant if you use istioClient directly

imikushin:
Looks much simpler now!

berndtj:
The code has been merged in

PR #615, opened 2018-09-06
adding updating docs script for releases

berndtj:
Docs can be update for each release with the following command:
```
./update-docs.sh v0.1.24
```


berndtj:
@pzmrzy It's not a bad idea.

PR #617, opened 2018-09-13
Minimal dispatch server based on knative

berndtj:
Can simply build and execute functions

* remove separate services/managers, go to a single dispatch
  server.
  - may still keep identity manager
* remove all obsolete charts
* remove old function manager (non-knative code)


config/300-function-manager.yaml
```diff
@@ -14,3 +14,12 @@ spec:
             image: github.com/vmware/dispatch/cmd/dispatch-server
             args:
             - function-manager
+            - --image-registry berndtj
```
neosab: No thanks ;). This can be ${KO_DOCKER_REPO}

config/300-function-manager.yaml
```diff
@@ -14,3 +14,12 @@ spec:
...
+            - --sourceroot /store
+            volumeMounts:
```
neosab: I thought this didn't work.

charts/dispatch/values.yaml
```diff
@@ -1,63 +1,58 @@
...
+  #  memory: 128Mi
+  #requests:
+  #  cpu: 100m
+  #  memory: 128Mi
+
+registry:
+  insecure: false
+  # Use https://index.docker.io/v1/ for dockerhub
+  url:
+  # Use 'echo -n "username" | base64' to generate this string
```
neosab: It will be nice if we base64 enc in helm

charts/dispatch/templates/_helpers.tpl
```diff
@@ -1,9 +1,26 @@
...
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "fullname" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
 {{- define "ingress_with_external_auth" -}}
-{{- $ingress_enabled := default .Values.global.ingress.enabled .Values.ingress.enabled -}}
+{{- $ingress_enabled := .Values.ingress.enabled -}}
```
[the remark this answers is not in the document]
berndtj: Thanks, yeah, I made a pass and removed all global refs.

config/300-function-manager.yaml
berndtj: Oh... yeah, I'm not sure I want to keep the config/*.yaml files.  Just one way to install.  But that's for another commit

config/300-function-manager.yaml
berndtj: Yeah, it doesn't :).  I'm reverting all my changes to config, but like I said, I'd like to delete it

charts/dispatch/values.yaml
berndtj: Yeah, I agree.  I didn't bother writing a helper to do so (or look to see if one already exists).

charts/dispatch/values.yaml
```diff
@@ -1,63 +1,58 @@
...
-    # Ignored, if rbac.create is true
-    serviceAccountName: default
-  resources: {}
-    # limits:
-    #  cpu: 100m
-    #  memory: 128Mi
-    #requests:
-    #  cpu: 100m
-    #  memory: 128Mi
+# Default values for function-manager.
```
neosab: Nit: Redundant Comments L4-L6

berndtj: Forgot this one.  Fixed now.  Should be all good to go

PR #618, opened 2018-09-14
Add istio based api-gateway

berndtj:
This merges in the istio based API gateway work by nickaashoek


tenczar:
I feel like I've missed something. I'm only seeing riff/lib and vendor code changes.

PR #620, opened 2018-09-17
Knative-based Dispatch images (image-manager)

zimengyang:
* Re-platform Dispatch image manager on Knative 
  * Map image as Knative `Build` CRD
  * Base-image will be a property in Dispatch Image. Not separate entity for now.
* CRUD Dispatch Image
* Rename `image-manager` to `images`
* Get rid of old `images` and `image_builder` pkg

TODO:
* Integrate with other Dispatch components
* Tests
* Make base-image constrained. by configmap or something else.


berndtj:
Great start!

pkg/images/backend/knhelpers.go
```diff
@@ -0,0 +1,192 @@
...
+			Steps: []corev1.Container{
+				corev1.Container{
+					Name:    string("write-system-package-files"),
+					Image:   string("vmware/photon2:20180424"),
+					Command: []string{"/bin/bash"},
+					Args:    []string{"-c", "echo -n ${SYSTEM_PACKAGES_CONTENT} | base64 -d > /workspace/${SYSTEM_PACKAGES_FILE}"},
+				},
+				corev1.Container{
+					Name:    string("write-package-files"),
+					Image:   string("vmware/photon2:20180424"),
```
neosab: Any specific reason we need this tag? and not `latest`

zimengyang: Will change to latest, forgot to change this while playing with knative. Thanks!

pkg/utils/knaming/meta.go
```diff
@@ -22,7 +22,9 @@ const (
 
 	KnTypeLabel = "knative.dev/type"
 
-	FunctionKnType = "function"
+	FunctionKnType  = "function"
+	BaseImageKnType = "baseimage"
```
neosab: AFAIK, only "function" or "app" are associated with `knative.dev/type` label. We can probably skip this for base image/ image & others.

pkg/images/backend/knhelpers.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// FromBaseImage produced Knative BuildTemplate from Dispatch BaseImage
+func FromBaseImage(baseimage *dapi.BaseImage) *knbuild.BuildTemplate {
+	if baseimage == nil {
+		return nil
+	}
+
+	// TODO: BaseImage -> BuildTemplate make this right and configurable
+	return &knbuild.BuildTemplate{
```
neosab: I am not sure if I fully understand the idea behind backing a `BaseImage` with a `BuildTemplate`. Seems like we could get away with a single instance `BuildTemplate` for a given version of dispatch.

neosab: In the legacy dispatch, `BaseImage` was an allowed golden image created by an admin and all images were its derivatives. If we just need that functionality, the set of allowed BaseImages could very well be defined in a config map.

zimengyang: Thanks! So, a better solution might be to treat Dispatch base-image as a parameter, which is a URL to the admin image. And store all allowed base image in a config map. The Knative Build (Dispatch Image) will only reference to the base_image URL. And we only need one BuildTemplate, Dispatch Image mapping to Knative Build CRD remains the same. Does this make sense?

zimengyang:
Thanks @tenczar ! I am still working on implementing those commented handlers/funcs, and I will clean them up later after pr finished.

pkg/images/knhandlers.go
```diff
@@ -0,0 +1,142 @@
...
+
+	dapi "github.com/vmware/dispatch/pkg/api/v1"
+	image "github.com/vmware/dispatch/pkg/images/gen/restapi/operations/image"
+	"github.com/vmware/dispatch/pkg/trace"
+	"github.com/vmware/dispatch/pkg/utils"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type knHandlers struct {
```
berndtj: There is nothing here that is knative specific outside of the backend.  I'm not sure that we need or should have a knative specific handler.

berndtj: I realize this is the pattern in place now, but we can change it.

PR #621, opened 2018-09-17
Fix frontend

pzmrzy:
1. Fix card in responsive modle for home page
2. Click hambuger button/overflow button to show first/second level navbar in reponsive mode
3. Disable overflow button on home/community page
4. Add scale.fix.js


berndtj:
Awesome.  Can you squash all your commits into a single one and add a commit message which describes what you did?

PR #630, opened 2018-09-25
Rename api-manager -> endpoints

berndtj:
* Follow existing patterns from functions and image
  - largely rewrote and validated that endpoints are managed
    correctly
  - use Meta object
* Remove dead-code (api-manager)
* Move utils.Constants to api/v1 to avoid circular dependencies


imikushin:
This generally LGTM

PR #632, opened 2018-09-26
Adding a new zookeeper-less controller used when no zookeeper configured

tenczar:
Modifies the NewController method to act as a factory method. Returns a JungleController (DefaultController with Zookeeper dependencies removed) when no zookeeper path is provided. Returns the DefaultController otherwise.


tenczar:
That's easy to do and it removes a lot of duplicate code. I'll make the change.

PR #633, opened 2018-09-27
add minio support

berndtj:
* picked up knative serving master
  - dep ensure was broken, so updated
* allow for object/minio function storage as well as file/nfs
  - removed dependent chart. User must install one or the other as prerequisite
* updated charts to allow for minio config
* updated functions for new "unstructured" build
* updated docs to reflect new installation


pkg/functions/backend/knative.go
```diff
@@ -152,18 +162,22 @@ func (h *knative) Update(ctx context.Context, function *dapi.Function) (*dapi.Fu
...
-		return "", errors.Wrapf(err, "getting knative route '%s'", serviceName)
+		return "", "", errors.Wrapf(err, "getting knative route '%s'", serviceName)
+	}
+
+	if h.ingressGatway == "" {
+		return "http://" + route.Status.DomainInternal, route.Status.DomainInternal, nil
 	}
 
-	return "http://" + route.Status.DomainInternal, nil
+	return "http://" + h.ingressGatway, route.Status.DomainInternal, nil
```
tenczar: Does the route.Status.DomainInternal no longer need the `http://` prefix?

pkg/functions/handlers.go
```diff
@@ -59,23 +63,76 @@ func ConfigureHandlers(api middleware.RoutableAPI, h Handlers) {
...
+	httpClient    *http.Client
+	namespace     string
+	imageRegistry string
+	storageConfig *config.StorageConfig
+	imagesClient  client.ImagesClient
 }
 
 // NewHandlers is the constructor for the function manager API knHandlers
-func NewHandlers(kubeconfPath, namespace, imageRegistry, sourceRootPath string, imagesClient client.ImagesClient) Handlers {
+func NewHandlers(kubeconfPath, namespace, imageRegistry, ingressGateway, buildImage string, storageConfig *config.StorageConfig, imagesClient client.ImagesClient) Handlers {
```
tenczar: This is getting a little long in the number of parameters. Would it make sense to combine some of these into an object that provides operations on related parameters? From here it looks like this would require significant changes to the defaultHandlers struct, so it may not be feasible.

PR #635, opened 2018-09-29
Make dispatch solo compact with funky knative

pzmrzy:


pkg/functions/docker/driver.go
```diff
@@ -248,7 +248,7 @@ func (d *Driver) findActiveContainer(ctx context.Context, functionID, functionRe
 // GetRunnable creates runnable representation of the function
 func (d *Driver) GetRunnable(e *functions.FunctionExecution) functions.Runnable {
 	return func(ctx functions.Context, in interface{}) (interface{}, error) {
-		bytesIn, _ := json.Marshal(functions.Message{Context: ctx, Payload: in})
+		bytesIn, _ := json.Marshal(in)
```
berndtj: I don't think it's this simple.  How are secrets passed?

pzmrzy: Then how about modify funky, to remove non-necessary field before call python image?
@tenczar

tenczar: For Knative Funky was modified to read the secrets from environment variables. This means that a function gets secrets passed to it at creation time. When a function is invoked, funky reads the secrets from the environment variables and adds them to the context that is passed to the language server.

tenczar:
Berndt is correct that this PR does not completely handle all of the changes necessary to make Dispatch Solo work with the Knative Funky. Function creation needs to be modified to set the secrets as environment variables. These variables need to take the form of `SECRETS=secret_name_1, secret_name_2, etc`. Then each secret value should be another variable labeled `d_secret_{secret name}`, e.g `d_secret_secret_name_1`, `d_secret_secret_name_2`

berndtj:
Also, this would be a GREAT opportunity to add e2e tests for solo (hopefully unmodified from the current e2e tests).

pzmrzy:
Do you think we can remove all the code not related to dispatch-solo?

berndtj:
I'm ok with cleaning up dead code for dispatch-solo, as long as we put it on a separate "solo" branch.

PR #651, opened 2018-10-11
Make e2e tests pass for images and functions

berndtj:
* update for different function cli behavior
* remove the status checks on baseimages (not currently supported)
* make updates work for images and functions
	- expose revision
* configure outbound connectivity for knative


e2e/tests/baseimages_update.yaml
```diff
@@ -0,0 +1,7 @@
+kind: BaseImage
+name: nodejs-base
+imageUrl: dispatchframework/nodejs-base:0.0.8
+language: python3
```
tenczar: This is weird. We are grabbing a nodejs base image but saying the language is python3. I think you just cut/pasted this from the images_update.yaml. Not sure why it was like this in that file...

berndtj: Yeah, that's what the test looked like originally, so I just left it.  But I too thought it odd.

berndtj:
I think we could use "skip" rather than commenting out... I should probably just do that.

e2e/tests/baseimages_update.yaml
kars7e: Sooo.... can we fix it in this PR? :-)

PR #673, opened 2018-10-11
Updates e2e to run against Dispatch solo

tenczar:
Testing dispatch solo e2e with a local concourse instance.


e2e/tests/variables.bash
```diff
@@ -2,9 +2,9 @@
 
 ### COMMON VARIABLES ###
 : ${DOCKER_REGISTRY:="dispatchframework"}
-: ${BASE_IMAGE_PYTHON3:="python3-base:0.0.9"}
-: ${BASE_IMAGE_NODEJS6:="nodejs-base:0.0.9"}
-: ${BASE_IMAGE_POWERSHELL:="powershell-base:0.0.10"}
-: ${BASE_IMAGE_JAVA:="java-base:0.0.10"}
+: ${BASE_IMAGE_PYTHON3:="python3-base:0.0.13"}
```
pzmrzy: Should it change to knative version?

pzmrzy:
when do `batch_create_images`, it run `dispatch create seed-image`, we should update embedded yaml file to new base images

e2e/tests/variables.bash
tenczar: We will have to do that with your change. The e2e tests will fail if we don't have your changes as well.

PR #679, opened 2018-10-12
Change example url

pzmrzy:
No response for `http://example.com` should be `http://www.example.com`

### DIFF
--- a/examples/python3/http_func.py
+++ b/examples/python3/http_func.py
@@ -3,5 +3,5 @@ import time
 import requests
 
 def handle(ctx, payload):
-    resp = requests.get("http://example.com")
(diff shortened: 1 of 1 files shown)


examples/python3/http_func.py
```diff
@@ -3,5 +3,5 @@
 import requests
 
 def handle(ctx, payload):
-    resp = requests.get("http://example.com")
+    resp = requests.get("http://www.example.com")
```
tenczar: Is this really necessary? It looks like `http://example.com` and `http://www.example.com` serve the same page.

pzmrzy: I just try again, it same, but on that day, I cannot access the url without www

Issue #122, opened 2018-01-12
CLI shell completion


imikushin:
It'd be nice to have shell (bash and zsh) tab-completion for dispatch CLI. 

- [ ] Bug
- [ ] Feature
- [x] Enhancement

## Detailed Description
CLI users would be able to hit TAB to complete or see possible completions (if there are several) of the current word in their CLI command. 

## Context
Typing commands will get easier and require less viewing of `--help`. This allows quicker (and cooler) demos and provides a more enjoyable user experience. 

## Complexity
- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [x] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

## Impact

- [ ] Low - Annoyance, but does not impact business or functinality
- [ ] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

seanhuxy:
It's cool, I am also interested in having this feature.

embano1:
Ah, I see where the issue is (thought cobra would have it out of the box)
https://github.com/spf13/cobra/issues/107

PR #681, opened 2018-10-14
CLI completion for bash added.

mshindo:
Fixes: #122

This is a minimal addition to support CLI completion for bash. More features (e.g., zsh support) can be implemented in the future.


mshindo:
Another PR #687 to add the example.

PR #685, opened 2018-10-16
Solo: Docker Backend for Event Driver

zimengyang:
* Add Docker Backend for Event Driver for Dispatch Solo
* Change driverclient to: 
  * make Host and Port configurable
  * include org in HTTP request
* Fix Ticker example

TODO:
* Expose event driver


pkg/event-manager/drivers/docker_backend.go
```diff
@@ -0,0 +1,165 @@
...
+		}, nil, nil, "")
+		if err != nil {
+			return errors.Wrap(err, "error creating container")
+		}
+
+		if err := d.dockerClient.ContainerStart(ctx, created.ID, types.ContainerStartOptions{}); err != nil {
+			return errors.Wrap(err, "error starting container")
+		}
+
+		driver.ContainerID = created.ID
```
kars7e: Take a look at https://github.com/vmware/dispatch/blob/master/pkg/functions/docker/driver.go#L95, you can use container labels to store the driver id to later [find and filter](https://github.com/vmware/dispatch/blob/master/pkg/functions/docker/driver.go#L214) by it. This will remove the need to store it in our db or expose the container ID in the API.

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -0,0 +1,170 @@
...
+
+	opts := filters.NewArgs()
+	opts.Add("label", labelEventDriverID+"="+driver.ID)
+	containers, err := d.dockerClient.ContainerList(ctx, types.ContainerListOptions{
+		Filters: opts,
+		All:     true,
+	})
+	if len(containers) != 1 || err != nil {
+		return errors.Wrap(err, "error getting container while deleting event driver")
+	}
```
tenczar: It might be useful to have a private `delete` method that just takes a container so we don't have to repeat this look up logic during an update.

zimengyang: I removed the container look up in Update(). Since we only need to deal with container in delete.

PR #686, opened 2018-10-17
Make edit command compact for function

pzmrzy:


pzmrzy:
@tenczar Yes it used, the `work-dir` flag not work for edit command, it work for function path, when you edit a function, `sourcepath` is related to `work-dir`

PR #687, opened 2018-10-18
Example added for CLI completion command.

mshindo:
The example was added to "completion" command for Linux and MacOS.

Signed-off-by: Motonori Shindo <motonori@shin.do>


mshindo:
Thanks for your prompt action!

PR #695, opened 2018-10-23
drivereventtype expose

### LINKED ISSUE #660: Event drivers should be exposable for push type events
### Expected behavior

Event drivers have an "expose" flag which means that they have an HTTP endpoint for event sources to POST events to.  This must be supported in Dispatch Solo as well.

### Impact

- [ ] High - Blocker

### Your Environment

pzmrzy:
Fix #660 
expose 80 port to a random available port 
First commit:  keep `entity_handler` just remove container then create a new one with port exposed
Second commit: Call `deploy` or `expose` by check `--expose` flag
@berndtj which one you think is better


e2e/tests/helpers.bash
```diff
@@ -341,4 +341,5 @@ cleanup() {
   delete_entities secret
   delete_entities eventdriver
   delete_entities eventdrivertype
+  sleep 5s
```
zimengyang: Karol pushed a pr which fixed the race condition issue of our e2e. Is this trying to solve the same problem?

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -95,32 +101,49 @@ func (d *dockerBackend) Deploy(ctx context.Context, driver *entities.Driver) err
...
+		freePort, err := GetFreePort()
+		if err != nil {
+			return errors.Wrap(err, "error get free port when creating container")
+		}
+
+		hostConfig := &container.HostConfig{}
+
+		if driver.Expose {
+			config.ExposedPorts = nat.PortSet{
+				"80/tcp": struct{}{},
```
zimengyang: To my understanding, all event driver will listen to port 80? This is not configurable to the event driver, am I right?

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -95,32 +101,49 @@ func (d *dockerBackend) Deploy(ctx context.Context, driver *entities.Driver) err
...
+
+		created, err := d.dockerClient.ContainerCreate(ctx, config, hostConfig, nil, "")
 		if err != nil {
 			return errors.Wrap(err, "error creating container")
 		}
 
 		if err := d.dockerClient.ContainerStart(ctx, created.ID, types.ContainerStartOptions{}); err != nil {
 			return errors.Wrap(err, "error starting container")
 		}
+		driver.URL = fmt.Sprintf("http://127.0.0.1:%d", freePort)
```
zimengyang: Where will driver.URL be referred later? Driver and solo will be running on same host, may i know why the URL is always 12.0.0.1 ?

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -77,11 +80,14 @@ func (d *dockerBackend) getDriverSecrets(ctx context.Context, driver *entities.D
...
-// Deploy deploys driver
-func (d *dockerBackend) Deploy(ctx context.Context, driver *entities.Driver) error {
-	log.Infof("Docker backend: deploying driver %v", driver.Name)
+// Expose deploys driver
+func (d *dockerBackend) Expose(ctx context.Context, driver *entities.Driver) error {
+	return nil
+}
 
-	// get driver secrets
+// Deploy exposes driver
```
zimengyang: nit: should this comment be "Deploy deploys driver"?

e2e/tests/helpers.bash
tenczar: Yeah, this can be removed at this point.

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -158,3 +181,18 @@ func (d *dockerBackend) Delete(ctx context.Context, driver *entities.Driver) err
 
 	return err
 }
+
+//GetFreePort return a free port
+func GetFreePort() (int, error) {
```
tenczar: Can you comment this function. I'm not sure how it gets a free port.

e2e/tests/helpers.bash
pzmrzy: Yes

pkg/event-manager/drivers/docker_backend.go
pzmrzy: Yes there seems no way to config port when create create a eventdriver

pkg/event-manager/drivers/docker_backend.go
pzmrzy: Just for print, when do `dispatch get event`, I think it should be only return the port, then dispatch cli will read the config and output full url. How do you think about this?

pkg/event-manager/drivers/docker_backend.go
pzmrzy: I also don't know, I just google a function, and it works

pkg/event-manager/drivers/docker_backend.go
```diff
@@ -78,11 +79,14 @@ func (d *dockerBackend) getDriverSecrets(ctx context.Context, driver *entities.D
 	return secrets, nil
 }
 
-// Deploy deploys driver
-func (d *dockerBackend) Deploy(ctx context.Context, driver *entities.Driver) error {
-	log.Infof("Docker backend: deploying driver %v", driver.Name)
+// Expose deploys driver
+func (d *dockerBackend) Expose(ctx context.Context, driver *entities.Driver) error {
```
tenczar: Is this function necessary? Why does it only return nil? There should at least be a comment about why it is unimplemented and personally I think it should throw some sort of UnimplementedError particularly if this method is just here to implement an interface.

tenczar: Then I would recommend making that comment and having this return an an unimplemented error

PR #698, opened 2018-10-25
Small changes to fix e2e errors in dispatch-solo

tenczar:


tenczar:
This change adds a small wait between deleting base images/images and reseeding them before testing updates. It also fixes a small issue in the API manager that used the incorrect idea.

PR #703, opened 2018-10-29
Add log output to file

### LINKED ISSUE #671: Dispatch Solo logs should be accessible via the CLI (UI?)
### Detailed Description

Debugging should not require access to the Dispatch Solo VM.  The logs should be accessible more easily via the API/CLI.

We may want to feature flag this as it could be a security issue.

### Complexity

pzmrzy:
Fix #671 
While server log output to console, it also output to a configurable file.
Then we can use `tail -f` to see log


pkg/dispatchserver/main.go
```diff
@@ -23,7 +23,13 @@ var dispatchConfigPath = ""
 
 // NewCLI creates cobra object for top-level Dispatch server
 func NewCLI(out io.Writer) *cobra.Command {
-	log.SetOutput(out)
+	logFile, err := os.OpenFile("./dispatch.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
tenczar: The path to this file should be configurable and it should be possible to disable logging to this file, which personally I think should be the default.

pzmrzy: Do you think we can add a flag, for example `--logfile filepath`

tenczar: I think that is reasonable.

kars7e:
This is ok to have, but I wouldn't use that to collect logs in OVA. The dispatch-server process is running as a service under systemd supervision, and because of that the logs are collected by journald. We could configure journal to log the file, and we can utilize systemd to collect logs when debugging. If we add additional logging to file, we need to manage that file (rotate it, secure it, etc.). This is something that systemd already does, so there is no need to duplicate that.

pzmrzy:
need more discussion, close for now

neosab:
> Then we can use `tail -f` to see log

FYI you can do the same with `journalctl -f -u dispatch-server`

PR #704, opened 2018-10-30
Updating CI scripts for knative CI

berndtj:
CI work is not complete, but scripts are getting there

* make targets in ci/Makefile work
	- make images
	- make cluster-create
	- make install-knative
	- make build-cli
	- make deploy-dispatch-server
	- etc

* updated knative dependencies (master as of a couple days ago)
* include and support internal docker repository by default
* updated readme


scripts/values.sh
```diff
@@ -1,26 +1,13 @@
 #! /bin/bash
```
kars7e: Do we still use this script? I can't remember what it was used for

berndtj: It's actually used as part of make images.  We could scrap it though.  Not sure how much value it provides

PR #705, opened 2018-10-30
Clean up Solo from not used components

kars7e:
- deleted all references to k8s
- deleted application, service components
- deleted charts
- deleted unused drivers


berndtj:
Are the tests flakey or did secrets break?

tenczar:
The tests seem to be flaky. In concourse it seems like the cleanup from one set of tests is overlapping with the batch create for the next set. I've added a short sleep to the end of cleanup, but I think we might need better locking when creating and deleting images. This may also be related to the dind used to run the tests...

PR #707, opened 2018-10-31
Fix image creation/deletion issue

kars7e:
There is a race condition with images where image might be in deleting
state, causing list API to not include it in the result, but also
not allowing to create a new one with the same name. This patch
attempts to fix it by always returning all images in the list, regardless of their
state, until the actual record is deleted from the database.

Also makes e2e number of retries and wait times configurable via global
variable (can still be specified in the retry function itself)

### DIFF
--- a/e2e/tests/helpers.bash
+++ b/e2e/tests/helpers.bash
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+: ${RET_NR:=30}
+: ${WAIT_T:=2}
+
 ### COMMON FUNCTIONS ###
 
 # retry_simple takes 3 args: command [interval (secs)] [attempts (num)]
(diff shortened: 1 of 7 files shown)


e2e/tests/helpers.bash
```diff
@@ -223,7 +228,26 @@ run_with_retry() {
...
+  echo ${1}
+  echo $output
+  return 1
+}
+
+# Run an arbitrary bash command until it succeeds
+#
+# usage: run_with_success <bash command> [retries] [sleep interval]
+#
+run_until_success() {
```
berndtj: I'm not sure how I would name this, but the current name makes me think it will just retry forever.

kars7e: Ha, good point. I will rename it to something more meaningful next time I'll be using it.

PR #709, opened 2018-11-01
Add hostname when list api endpoint

### LINKED ISSUE #609: The CLI should include the hostname when listing API endpoints
## Feature Request

Currently the CLI only lists out the path associated with an API endpoint.  This requires the user to know the hostname of the API gateway.  It's very likely that the user does not know this.  While the user can inspect the context/config to get the dispatch API hostname, this is different than the API gateway hostname.

### Complexity

pzmrzy:
Fix #609 

Should add api gateway port in dispatch client config, current just return 8081


pkg/dispatchcli/cmd/get_api.go
```diff
@@ -85,9 +85,17 @@ func formatAPIOutput(out io.Writer, list bool, apis []v1.API) error {
...
 		})
 	}
 	table.Render()
 	return nil
 }
+
+func addHostName(uris []string) []string {
+	var newUris []string
+	for _, uri := range uris {
+		newUris = append(newUris, dispatchConfig.Host + ":8081" + uri)
```
tenczar: How do we display the http endpoint versus the https endpoint? They will certainly have different ports. I think this requires a little more thought in exactly how we fulfill this requirement. I would imagine adding a new command to the api command that gets the base url and displaying those for http and https.

pkg/dispatchcli/cmd/get_api.go
```diff
@@ -85,9 +85,25 @@ func formatAPIOutput(out io.Writer, list bool, apis []v1.API) error {
...
 	table.Render()
 	return nil
 }
+
+func addHostName(uris []string, protocals []string) []string {
+	var newUris []string
+	for _, uri := range uris {
+		for _, protocal := range protocals {
+			if protocal == "http" {
+				newUris = append(newUris, "http://"+dispatchConfig.Host+":8081"+uri)
```
tenczar: This really should be able to be simplified to
```
for _, protocol := range protocols {
    newUris = append(newUris, fmt.Sprintf("%s://%s:%s%s", protocol, dispatchConfig.Host, portMap[protocol], uri))
}
```
where portMap is some local or package variable that contains a map of the protocol to the API gateway port for that protocol. We really do need a way of adding the correct port information here. Hard coding 8081 or 8444 won't work.

pzmrzy: I find api ports in dispatchConfig, just update to: if the port is 0 then use default port

PR #710, opened 2018-11-02
change disk controller in ova

berndtj:


kars7e:
Tested on VMC, works correctly. on Fusion there is another issue with ovftool, but it's unrelated.

PR #711, opened 2018-11-05
Update OVA to use photon 2.0

kars7e:
* move ova infra to top-level directory from scripts subdirectory
* Use dispatch image for building OVA based on photon 2.0
* Update scripts to work properly on photon 2.0
* Fix ovfenv binary to work properly in fusion
* Fix minor bugs causing errors in OVA logs

photon 2.0-based OVA is ~30% smaller than the previous one

Tested on Fusion 11.0 and VMC


ova/bootable/build-app.sh
```diff
@@ -34,17 +34,23 @@ chage -I -1 -m 0 -M 99999 -E -1 root
 
 log3 "configuring ${brprpl}UTC${reset} timezone"
 ln --force --symbolic /usr/share/zoneinfo/UTC /etc/localtime
-log3 "configuring ${brprpl}en_US.UTF-8${reset} locale"
-/usr/bin/touch /etc/locale.conf
-/bin/echo "LANG=en_US.UTF-8" > /etc/locale.conf
-/sbin/locale-gen.sh
+
+## TODO: This causes issues on photon 2.0, debug or remove permanently when root cause
```
zjs: Should we file as issue to track this? (I'm unsure of the Dispatch conventions here, but that may make it easier to associate this with VIC; they'll be looking at Photon 2.0 for 1.5.)

kars7e:
Thanks Zach!

ova/bootable/build-app.sh
kars7e: Sounds good, thanks for reminding me!

DanielXiao: You can try to install glibc-i18n package.

PR #713, opened 2018-11-05
add dispatch CLI to the OVA

kars7e:
* Add Dispatch CLI and its config to the OVA
* open port 443 for API Gateway
* set the api-endpoint to docker0 ip address for event drivers

### DIFF
--- a/ova/scripts/systemd/units/dispatch-server.service
+++ b/ova/scripts/systemd/units/dispatch-server.service
@@ -11,7 +11,7 @@ Restart=on-failure
 RestartSec=10
 
 WorkingDirectory=/var/lib/dispatch
-ExecStart=/usr/bin/dispatch-server --host 0.0.0.0 --enable-tls --tls-certificate /storage/data/certs/server.crt --tls-certificate-key /storage/data/certs/server.key
+ExecStart=/usr/bin/dispatch-server --events-gateway 172.17.0.1:8080 --host 0.0.0.0 --gateway-tls-port 443 --enable-tls --tls-certificate /storage/data/certs/server.crt --tls-certificate-key /storage/data/certs/server.key
 
 [Install]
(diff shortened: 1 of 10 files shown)


ova/scripts/systemd/units/dispatch-server.service
```diff
@@ -11,7 +11,7 @@ Restart=on-failure
 RestartSec=10
 
 WorkingDirectory=/var/lib/dispatch
-ExecStart=/usr/bin/dispatch-server --host 0.0.0.0 --enable-tls --tls-certificate /storage/data/certs/server.crt --tls-certificate-key /storage/data/certs/server.key
+ExecStart=/usr/bin/dispatch-server --api-endpoint 172.17.0.1:8080 --host 0.0.0.0 --gateway-tls-port 443 --enable-tls --tls-certificate /storage/data/certs/server.crt --tls-certificate-key /storage/data/certs/server.key
```
berndtj: Is that api-endpoint right?  Will it always be addressible at that IP?

tenczar: At the very least I think it should be given a descriptive variable name so people know what that IP is supposed to be.

kars7e: @berndtj yes, this is the IP of docker0 interface, always set to this IP in photon OVA, reachable from all docker containers. 
@tenczar can you propose some name? I'm honestly at loss here. this is the endpoint that is reachable by event drivers (running as containers).

tenczar: Could we call it DOCKER_0_BRIDGE_IP or something similar?

kars7e: Renamed to events-gateway. Thanks @berndtj !

Issue #668, opened 2018-10-11
Let's Encrypt/cert-bot integration for trusted certificates


berndtj:
### Detailed Description

If the user configures Dispatch Solo with a domain, use Let's Encrypt to generate valid certificates for all Dispatch Solo endpoints.

### Complexity

- [ ] Low - Simple enhancement or bug fix, no architectural changes or refactoring
- [ ] Medium - Change requires some thought, but is relatively isolated
- [ ] High - Significant architectural change or large refactor

PR #715, opened 2018-11-07
Lets Encrypt Integration

pzmrzy:
When first time run the server, it will automatic get certificate from Lets Encrypt
For staging environment:
`sudo ./bin/dispatch-server-linux --host 0.0.0.0 --enable-tls --lets-encrypt --domain <domain>`
For production environment:
`sudo ./bin/dispatch-server-linux --host 0.0.0.0 --enable-tls --lets-encrypt --domain <domain> --production`
#668

### DIFF
--- a/pkg/dispatchserver/config.go
+++ b/pkg/dispatchserver/config.go
@@ -36,8 +36,11 @@ type serverConfig struct {
 	DisableHTTP       bool   `mapstructure:"disable-http" json:"disable-http"`
 	TLSPort           int    `mapstructure:"tls-port" json:"tls-port"`
 	EnableTLS         bool   `mapstructure:"enable-tls" json:"enable-tls"`
+	LetsEncrypt       bool   `mapstructure:"lets-encrypt" json:"lets-encrypt"`
+	Production        bool   `mapstructure:"production" json:"production"`
 	TLSCertificate    string `mapstructure:"tls-certificate" json:"tls-certificate"`
 	TLSCertificateKey string `mapstructure:"tls-certificate-key" json:"tls-certificate-key"`
+	Domain            string `mapstructure:"domain" json:"domain"`
 
 	Tracer string `mapstructure:"tracer" json:"tracer"`
 	Debug  bool   `mapstructure:"debug" json:"debug"`
@@ -87,6 +90,9 @@ func configGlobalFlags(flags *pflag.FlagSet) {
 	flags.String("tls-certificate", "", "Path to the certificate file")
 	flags.String("tls-certificate-key", "", "Path to the certificate private key")
 	flags.Bool("enable-tls", false, "Enable TLS (HTTPS) listener.")
+	flags.Bool("lets-encrypt", false, "Use Let's Encrypt staging to generate certificate")
+	flags.Bool("production", false, "Use Let's Encrypt production server")
+	flags.String("domain", "", "Server domain name")
 
 	flags.String("tracer", "", "OpenTracing-compatible Tracer URL")
 	flags.Bool("debug", false, "Enable debugging logs")
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -22,6 +22,16 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/netutil"
+	"os"
+	"strings"
+	"github.com/vmware/dispatch/pkg/utils"
+	"path"
+	"crypto"
+	"crypto/rsa"
+	"crypto/rand"
+	"encoding/pem"
+	"golang.org/x/crypto/acme"
+	"crypto/x509/pkix"
 )
 
 const (
@@ -119,6 +129,9 @@ type Server struct {
 	TLSReadTimeout time.Duration
 	// TLSWriteTimeout sets the maximum duration before timing out write of the response for HTTPS connections.
 	TLSWriteTimeout time.Duration
+	LetsEncrypt     bool
+	Production      bool
+	Domain          string
 	httpsListener   net.Listener
 
 	handler         http.Handler
@@ -139,6 +152,16 @@ func (s *Server) hasScheme(scheme string) bool {
 		}
 	}
 	return false
+}
+
+func (s *Server) existCertificate(uDomain string) bool {
+	if _, err := os.Stat("./lets_encrypt/"+uDomain+".key"); os.IsNotExist(err) {
+		return false
+	}
+	if _, err := os.Stat("./lets_encrypt/"+uDomain+".crt"); os.IsNotExist(err) {
+		return false
+	}
+	return true
 }
 
 // Serve the api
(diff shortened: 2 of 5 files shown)


pkg/dispatchserver/config.go
```diff
@@ -36,8 +36,11 @@ type serverConfig struct {
 	DisableHTTP       bool   `mapstructure:"disable-http" json:"disable-http"`
 	TLSPort           int    `mapstructure:"tls-port" json:"tls-port"`
 	EnableTLS         bool   `mapstructure:"enable-tls" json:"enable-tls"`
+	LetsEncrypt       bool   `mapstructure:"lets-encrypt" json:"lets-encrypt"`
+	Production        bool   `mapstructure:"production" json:"production"`
```
kars7e: IMHO you can drop this flag. staging environment is used for testing, e.g. when you are developing automation for that. you could have an internal constant, or a config flag to set the let's encrypt server, but production server should be the default.

pkg/dispatchserver/config.go
```diff
@@ -36,8 +36,11 @@ type serverConfig struct {
...
 	TLSCertificate    string `mapstructure:"tls-certificate" json:"tls-certificate"`
 	TLSCertificateKey string `mapstructure:"tls-certificate-key" json:"tls-certificate-key"`
+	Domain            string `mapstructure:"domain" json:"domain"`
```
kars7e: If the domain is needed, how about we merge lets-encrypt and this flag and have only one, `lets-encrypt-domain` flag? If it's empty, we disable lets encrypt. if it's set, we use the value as the domain.

pkg/utils/constants.go
```diff
@@ -52,3 +52,18 @@ const ServiceAccountKind = "ServiceAccount"
 
 // OrganizationKind a constant representing the kind of the Organization Model
 const OrganizationKind = "Organization"
+
+// LetsEncryptStaging a constant representing Let's Encrypt staging server url
+const LetsEncryptStaging = "https://acme-staging.api.letsencrypt.org/directory"
```
kars7e: these constants look http-specific, move them to `http` package. In general, avoid `utils` package if possible.

pkg/http/server.go
```diff
@@ -185,6 +199,24 @@ func (s *Server) Serve() (err error) {
...
+		if s.LetsEncrypt {
+			if s.Domain == "" {
+				return errors.New("Need a domain name, use --domain <Domain Name>")
+			}
+			if !s.Production {
+				log.Warnf("Staging environment, use --production for production environment")
+			}
+			uDomain := strings.Replace(s.Domain, ".", "_", -1)
+			if s.existCertificate(uDomain) {
+				s.TLSCertificate = "./lets_encrypt/" + uDomain + ".crt"
```
kars7e: `./lets_encrypt/` should be a constant, you are using it in multiple places.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+
+	// Attempt authorization
+	auth, err := c.Authorize(context.TODO(), s.Domain)
+	if err != nil {
+		return err
+	}
+	var challenge *acme.Challenge
+	for _, c := range auth.Challenges {
+		if c.Type == "http-01" {
+			challenge = c
```
kars7e: I'm guessing you are interested in the first occurrence of a challenge of type `http-01`, if so you could add `break` here to break out of the loop as soon as you find it.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+	log.Infof("Registration succeeded")
+
+	// Attempt authorization
+	auth, err := c.Authorize(context.TODO(), s.Domain)
+	if err != nil {
+		return err
+	}
+	var challenge *acme.Challenge
+	for _, c := range auth.Challenges {
+		if c.Type == "http-01" {
```
kars7e: http-01 means a chalange over unauthenticated HTTP and port 80, which means people who want to listen only on port 443 will have to open one more port in the firewall. let's talk more about this offline.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+	if _, err := os.Stat("./lets_encrypt/" + uDomain + ".key"); os.IsNotExist(err) {
+		return false
+	}
+	if _, err := os.Stat("./lets_encrypt/" + uDomain + ".crt"); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func (s *Server) getLetsEncrypt(uDomain string) error {
```
kars7e: this function is really long, let's break it down into smaller pieces. take a look at `crypto.go`, maybe we can share some logic with it.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+	}
+
+	log.Infof("Complete!")
+
+	s.TLSCertificateKey = fmt.Sprintf("%s.key", path.Join(certDirectory, uDomain))
+	s.TLSCertificate = path.Join(certDirectory, fmt.Sprintf("%s.crt", uDomain))
+	return nil
+}
+
+func newKey(filename string) (crypto.Signer, error) {
```
kars7e: `crypto.go` already generates private key, maybe we could reuse that (also it uses ecdsa, which has the advantage of a smaller key size).

pkg/utils/constants.go
```diff
@@ -52,3 +52,18 @@ const ServiceAccountKind = "ServiceAccount"
...
+
+// LetsEncryptProduction a constant representing Let's Encrypt production server url
+const LetsEncryptProduction = "https://acme-v01.api.letsencrypt.org/directory"
+
+// KeyLength a constant representing the length of TLS Key
+const KeyLength = 2048
+
+// KeyType a constant representing the length of TLS Key Type
```
tenczar: Comment is a copy of the previous comment.

pkg/http/server.go
tenczar: I would go a step further and say this should be a configurable parameter. If we are going to be writing files we should give the end user a choice of where those files go.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+	}
+	if _, err := os.Stat("./lets_encrypt/" + uDomain + ".crt"); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func (s *Server) getLetsEncrypt(uDomain string) error {
+	letsEncryptURL := utils.LetsEncryptStaging
+	certDirectory := "./lets_encrypt"
```
tenczar: I would move this below the following `if` block. Put all of the initialization of `letsEncryptURL` in one place.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
...
+
+	// Create the ACME client
+	c := &acme.Client{
+		Key:          aKey,
+		DirectoryURL: letsEncryptURL,
+	}
+
+	// Begin with registration
+	log.Infof("Attempting registration...")
+	_, err = c.Register(context.TODO(), nil, func(string) bool { return true })
```
tenczar: you can do `_, err := c.Register(context.TODO(), nil, acme.AcceptTOS)` to make this a little more clear to a human reader.

pkg/http/server.go
```diff
@@ -381,3 +413,164 @@ func (s *Server) HTTPSURL() string {
 	}
 	return ""
 }
+
+func (s *Server) existCertificate(uDomain string) bool {
```
berndtj: existingCertificate

PR #716, opened 2018-11-07
change events gateway flag in e2e

kars7e:
#713 changed the flag name, but there was one place where it wasn't changed (e2e tests). This PR fixes that.

Also, should we run e2e on every PR in solo, given the tests run much shorter now? It's around ~14 minutes now (we could run automatically if there is a change in *.go or e2e files)


kars7e:
I guess I forgot about the test run on merge (which btw helped me see the bug this PR fixed). I guess it's ok if we break the e2e from time to time as long as we can quickly act upon that. I'm ok with status quo then

PR #741, opened 2018-11-15
Fix ova config

### LINKED ISSUE #738: Invalid G/W port in API path
## Bug Report

I am seeing an invalid port in the API path. This must be the g/w's tls port 443

root@null [ ~/dispatch ]# dispatch get api
     NAME    | FUNCTION | PROTOCOL | METHOD | DOMAIN |                 PATH                  | AUTH | STATUS | ENABLED
------------------------------------------------------------------------------------------------------------------------
  post-hello | hello-py | https    | POST   |        | https://localhost:8444/dispatch/hello |      | READY  | true
------------------------------------------------------------------------------------------------------------------------

### Impact

- [X] Medium - Issue can be worked around, but is causing pain
- [ ] High - Blocker

### Your Environment

pzmrzy:
Fix #738 
fix docker config when build ova locally
change ui destination to `/var/www`

### DIFF
--- a/ova/build.sh
+++ b/ova/build.sh
@@ -23,6 +23,9 @@ ROOT_WORK_DIR="/go/src/github.com/vmware/dispatch/"
 ROOT_INSTALLER_DIR="${ROOT_DIR}/ova"
 ROOT_INSTALLER_WORK_DIR="${ROOT_WORK_DIR}/ova"
 
+ROOT_UI_DIR="$GOPATH/src/github.com/vmware/dispatch-ui-binaries"
+ROOT_WORK_UI_DIR="/go/src/github.com/vmware/dispatch-ui-binaries"
+
 CI_IMAGE="dispatchframework/ova-builder:latest"
 
 TAG=${TAG:-$(git describe --tags --dirty)}
@@ -46,6 +49,7 @@ if [ "$step" == "ova-dev" ]; then
   docker run -it --rm --privileged -v /dev:/dev \
     -v "${ROOT_DIR}/:/${ROOT_WORK_DIR}/":ro \
     -v "${ROOT_INSTALLER_DIR}/bin/:/${ROOT_INSTALLER_WORK_DIR}/bin/" \
+    -v "${ROOT_UI_DIR}":"${ROOT_WORK_UI_DIR}" \
     -v /var/run/docker.sock:/var/run/docker.sock \
     -e DEBUG="${DEBUG}" \
     -e TAG="${TAG}" \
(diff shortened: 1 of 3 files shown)


ova/build.sh
```diff
@@ -23,6 +23,9 @@ ROOT_WORK_DIR="/go/src/github.com/vmware/dispatch/"
 ROOT_INSTALLER_DIR="${ROOT_DIR}/ova"
 ROOT_INSTALLER_WORK_DIR="${ROOT_WORK_DIR}/ova"
 
+ROOT_UI_DIR="$GOPATH/src/github.com/vmware/dispatch-ui-binaries"
```
berndtj: Does this require the user to have checked out the UI explicitly or is it done automatically earlier.

pzmrzy: in pipeline it automatically download the repo, and compile. But if build local, the use must have the ui repo in the right place.
Or I can add some script, to automatically clone the ui repo and build.

berndtj: Ask @kars7e's opinion also, but I would expect it to pull the repo and/or checkout origin/master when building.  Maybe allow an override with an ENV var, but otherwise the UI will always be stale when building locally unless you remember to update the UI repo before building.

berndtj: I'm not sure I like depending on the code being checked out.

kars7e: I agree with @berndtj, This actually does not fix anything, instead, it breaks stuff. what if `$GOPATH/src/github.com/vmware/dispatch-ui-binaries` does not exist? you should not depend on it.

Eventually, we want to have a pipeline that builds using versioned artifacts, but before we get there we need to ensure what we have today is functioning.

ova/build.sh
```diff
@@ -23,6 +23,9 @@ ROOT_WORK_DIR="/go/src/github.com/vmware/dispatch/"
...
+ROOT_WORK_UI_DIR="/go/src/github.com/vmware/dispatch-ui-binaries"
```
kars7e: how about just `ui` directory in dispatch directory?
`ROOT_WORK_UI_DIR="${ROOT_DIR}/ova`
`ROOT_DIR` is already mounted in local docker, and you keep everything inside dispatch.
Using `/go/src/github.com/vmware/dispatch-ui-binaries` kind of suggests it's a go-dependent directory.

ova/build.sh
```diff
@@ -41,11 +44,20 @@ step=$1; shift
 [ ! "$step" == "ova-ci" ] || [ ! "$step" == "ova-dev" ] || usage
 
 echo "--------------------------------------------------"
+if [ ! -d ${ROOT_UI_DIR} ]; then
+  export LATEST=$(curl -s https://api.github.com/repos/dispatchframework/dispatch-ui/releases/latest | jq -r .name)
```
kars7e: add `echo` command saying "UI directory does not exist, downloading UI release..."

PR #746, opened 2018-11-20
[solo] quickstart VMware fusion version

zimengyang:


kars7e:
Why do we need version 11.0.1? Could we have an issue referencing the problems with versions < 11.0.1?

zimengyang:
closed, tracked in issue #747

PR #749, opened 2018-11-20
Add api gateway https port in quick start

pzmrzy:
Add api gateway https port in quick start
There should no dash in config file

### DIFF
--- a/ova/configs/config.json
+++ b/ova/configs/config.json
@@ -6,7 +6,7 @@
             "scheme": "http",
             "organization": "dispatch",
             "insecure": true,
-            "api-https-port": 443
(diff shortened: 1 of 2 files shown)


ova/configs/config.json
```diff
@@ -6,7 +6,7 @@
             "scheme": "http",
             "organization": "dispatch",
             "insecure": true,
-            "api-https-port": 443
+            "apihttpsport": 443
```
berndtj: is this key correct?  I can't find reference to it in the source

pzmrzy: Yes, it must no dash
@zimengyang find this

PR #774, opened 2018-11-28
Add vcenter events guide

kars7e:
* Add an example of running hardening script with vSphere events.
* Fix a typo in local config causing eventsgateway setting  to be ignored.
* remove release unrelated to Solo.


docs/_examples/vcenter-events.md
```diff
@@ -0,0 +1,223 @@
...
+
+1. Create a JSON file with secret contents:
+
+```bash
+cat << EOF > vsphere.json
+{
+  "host": "<vCenter host URL, e.g. https://vcenter.example.com>",
+  "username": "<username>",
+  "password": "<password>"
+  "vcenterurl": "<full vCenter URL, e.g. username:password@vcenter.example.com:443>"
```
berndtj: Do we still have the password issues?  Also, at some point we should probably not require the URL since we already have all of the pieces which make up the URL.

docs/_examples/vcenter-events.md
```diff
@@ -0,0 +1,223 @@
...
+``` 
+
+Replace `myvm` with the name of your vm. 
+
+# Create an event driver
+In order to receive vSphere events we need to create an event driver.
+
+1. Create a vcenter event driver type which registers the docker image implementing the driver:
+```
+$ dispatch create event-driver-type vcenter dispatchframework/dispatch-events-vcenter:solo-auth
```
berndtj: We should figure out some tagging pattern that's consistent across the event drivers.

PR #784, opened 2018-12-04
Add WithURL opt to driverclient

neosab:
To make our event drivers compatible with knative container source, they must be able to send events to an HTTP URL. None of the current driverclient options allow that.

### DIFF
--- a/pkg/events/driverclient/http.go
+++ b/pkg/events/driverclient/http.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


kars7e:
Seems like using URL works for every case, maybe we could remove others and just leave WithURL?

neosab:
> Seems like using URL works for every case, maybe we could remove others and just leave WithURL?

@kars7e I am fine with that. With all other options, the events endpoint path `v1/events/ingest` was implicitly added by the driverclient. If we are going ahead with only `WithURL` opt then we just need to document the endpoint for driver implementors.

kars7e:
> @kars7e I am fine with that. With all other options, the events endpoint path `v1/events/ingest` was implicitly added by the driverclient. If we are going ahead with only `WithURL` opt then we just need to document the endpoint for driver implementors.

Great, let's merge it as is, if we remove the current options we need to update all the drivers and that's a pain currently, let's wait with that until we have a CI/build infra for all of them in place.

PR #794, opened 2018-12-24
Update README.md

james65535:
Missing hyphen for instruction to create base-image.


kars7e:
Both `baseimage` and `base-image` should work, are you seeing an issue when running it without the hyphen?

tompscanlan:
@kars7e I've seen an error when running without the hyphen in version v0.1.22-43-g66427ae5

PR #797, opened 2018-12-28
Fix docker registry and minio endpoints

kars7e:
Values for Dispatch helm chart use hard-coded values for docker registry
and minio endpoints. The values.yaml should reflect that.


tompscanlan:
Looks appropriate to me.  This recently popped up in running through note to install on cloud pks.  I needed to adjust the values.yaml directly over generating it with this script

